PR #1, opened 2018-04-25
fix(Socket): fix parsing of packet name and length

Lange:
This also fixes an issue with infinite recursion. I'm unsure what the purpose of the deleted `if` block was.

### DIFF
--- a/src/lib/atemSocket.ts
+++ b/src/lib/atemSocket.ts
@@ -138,8 +138,8 @@ export class AtemSocket extends EventEmitter {
 	}
 
 	private _parseCommand (buffer: Buffer, packetId?: number) {
-		let length = buffer.readUInt8(0)
-		let name = buffer.readUInt16BE(4)
+		let length = buffer.readUInt16BE(0)
+		let name = buffer.toString('utf8', 4, 8)
 
 		// this.log('COMMAND', `${name}(${length})`, buffer.slice(0, length))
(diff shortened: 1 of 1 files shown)


baltedewit:
If there are multiple commands in a single packet the if block is supposed to parse the next command.

Lange:
Ahh ok, currently it causes an infinite loop though because the length of `buffer` isn't modified anywhere.

Lange:
To properly implement this recursion, I think we'd need to do what Julian does in the C# library, and create something which pops bytes off of the buffer as it reads it, rather than what we do now which never mutates the buffer.

baltedewit:
> because the length of buffer isn't modified anywhere.

`buf.slice(offset)` returns a buffer of a new length to the next recursive call. This should theoretically reduce `buf.length` by exactly one command. There might be the possibility of `length` being 0, in which case we _will_ enter an infinite loop so that should get a fix. I don't see any issues with using `buf.slice(offset)` though.

Lange:
Oh my bad, yes I missed that there is a `buffer.slice(length)` happening in the recursive call. Regardless, it is currently looping infinitely. I'll work on it.

Lange:
Ah of course, the problem must be in the parsing of `length`. It's currently always returning zero. This also explains why this functionality broke when it did. Fix incoming shortly.

Lange:
There we go, this should be good to merge now.

PR #3, opened 2018-04-25
wip(API): implement Partial-based API

Lange:
> ❗ This PR is a work-in-progress. It should not be merged.

This is a proof-of-concept implementation of a new API structure which is based around the idea of supplying Partial state subtree objects as the primary argument in most API methods.

This approach allows us to coalesce dozens of separate API methods into one, greatly reducing the API surface, complexity, and maintenance cost.

There are further simplifications to make still. These simplifications are outlined in these TODO comments:

```ts
updateMediaPlayer (newProps: Partial<MediaPlayer>, player = 0) {
	let command = new Commands.MediaPlayerStatusCommand()
	command.mediaPlayerId = player

	/* TODO(Lange - 2018/04/25): This feels messy. We're on the right track, but need to simplify further.
	 * Specifically, it'd be nice if this updateMediaPlayer method didn't
	 * need to know about the shape of this.state.
	 */
	command.properties = {
		...this.state.media.players[player],
		...newProps
	}

	/* TODO(Lange - 2018/04/25): This also seems like it could be further simplified and automated.
	 * It'd be neat if there was a standard spec for calculating flags from a Partial of properties,
	 * which all commands adhered to. That way we could define calcFlags in one place, instead
	 * of each command needing to define it individually.
	 */
	command.flag = command.calcFlags(newProps)

	return this.sendCommand(command)
}
```

At this time, not all of the commands have been updated to this new structure. `Media/MediaPlayerStatusCommand.ts` and `MixEffects/Transition/TransitionDipCommand.ts` are two files which *have* been updated, and are good examples to look at to get a feel for what these changes do.

This seems like a strong start, but will likely need a bit more constructive feedback before it is ready to be fully implemented.


baltedewit:
I've tried to make an abstraction for setting properties s.t. the `atem.ts` file is simplified to this:

```
updateMediaPlayer (newProps: Partial<MediaPlayer>, player = 0) {
	let command = new Commands.MediaPlayerStatusCommand()
	command.mediaPlayerId = player
	command.updateProps(newProps)
	return this.sendCommand(command)
}
```

The full set of changes can be seen here:
https://gist.github.com/baltedewit/70688d87759d1d13d458501a0e03b09b

Lange:
Nice, that looks great to me. I'm just now heading to bed but feel free to add on to this branch if you like.

PR #5, opened 2018-04-27
feat(Inputs): implement InPr (read) and CInL (write) commands

Lange:
I've done a good amount of spot-checking on the deserialization, and it seems to be correct.

I have not tested the serialization at all, we'll probably want to double-check that.

### DIFF
--- a/src/atem.ts
+++ b/src/atem.ts
@@ -13,6 +13,7 @@ import {
 	TransitionProperties,
 	WipeTransitionSettings
 } from './state/video'
+import {InputChannel} from "./state/input";
 
 export interface AtemOptions {
 	localPort?: number,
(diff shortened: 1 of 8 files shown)


baltedewit:
I've tested the serialization and it's seemingly in line with both LibAtem and the official clients now, however I am not yet seeing parameters get updated.

Lange:
There seems to be a bug where the first write packet we send doesn't work,
regardless of the command. All subsequent write packets seem to work after
that though.

I haven't tested this theory with CInL yet, but it's what I experienced
with CSBP.

baltedewit:
Ah, yeah, it's the same with `CInL`. Looks like the Atem expects the first packetId to be 1 and not 0. I will merge this and push a fix for the packet.

PR #6, opened 2018-04-27
feat(SuperSource): implement SSBP (read) and CSBP (write) commands

Lange:
Similar to #5, I've done a good amount of testing on the deserialization, but I've not tested the serialization at all yet. We will need to do that before merging.


Lange:
CSBP has been fully tested and is good to go now 👌

Lange:
Here's an early screenshot of the alternative Super Source UI I'm working on:

image

Everything shown in this image works in both directions (read and write), in realtime, and is how I tested this PR.

PR #8, opened 2018-05-06
feat(Upstream Keyers)

Lange:
This PR is a big one: it fully implements Upstream Keyers.

As of it right now it is _completely_ untested, both for serialization and deserialization. Much testing needs to be done before merging.


Lange:
I've force-pushed fixes to the fill/cut source commands, they should work now.

Lange:
Whoops, that force-push failed and I hadn't noticed. _Now_ it should be good.

baltedewit:
I've done some more checking, and everything seems to be correct apart from the DVE Border settings and one weird thing where the DVE mask in the state becomes `value - 1`.

Lange:
Do you have a repro for the issue with DVE Border settings? I checked locally and it _seems_ like things are correct, but I could be overlooking something.

I've also not yet reproduced the described issue with DVE masks.

baltedewit:
The issue with the DVE masks is probably due to a rounding error on the ATEM's side, the parsing of the protocol on our side seems correct. (Also, it is only reproducible with significantly small values)

I don't see any other issues with merging this.

Lange:
Ah yeah the UI does rounding, so it's not always the same as what the
protocol sends.

PR #11, opened 2018-05-16
feat: Downstream Keyer Settings

mint-dewit:
This introduces the low level commands and high level methods for reading and writing Downstream Keyer Settings.

As of now this is still _untested_.


mint-dewit:
Tested and pushed fixes.

PR #17, opened 2018-08-06
wip: implement media protocol

Lange:
**DO NOT MERGE.** This PR is a work-in-progress.

This PR implements the media transfer protocol for uploading of clips, stills, and audio.


Lange:
This is blocked on #19.

Currently, the act of just reading frames from disk can block the main thread long enough for the ATEM to think we have died, and kill our connection. Even using read streams is not adequate, because once you have about 4 concurrent streams, the socket will still get blocked.

Lange:
Audio upload does not seem to work yet.

baltedewit:
The code for this currently look quite healthy but in my testing I am still seeing two blocking issues:

- Random mid transfer hangs.
- A bunch of 0 error codes, if I continue the transfer it loses the lock and the atem send more 0 error codes.

Now, the 0 error codes seem like they can be ignored, as they are intended for a transfer that has already finished. I have to investigate whether they have any effect on the actual still pool, but I hope I can resolve this and it could just be a side effect of issues described below.

The random mid transfer hangs seem to be related to a IPC messages failing to send. It could be that the amount of data being transferred through the IPC channel is too high?
Interestingly this coincides with multiple packets being resent (for reasons unknown to me this happens faster than the interval for resending runs), and the socket stops receiving ATEM pings even though the ATEM does send these out. A console log on an interval printing to stdout shows that the actual processing is still running though.

If I do manage to send out more than 20 or so files, I am seeing deteriorating upload speeds. (But no specific reason as to why we are seeing that.)

baltedewit:
I managed to resolve my issues with the mid transfer hangs. This ended up being the IPC channel getting saturated and its `send()` method returning `false` causing an immediate retry. This retry however just caused more congestion. Disabling resend upon the method returning `false` resolved my issue.

I can now upload about 60 files back to back, and then the transfers will stop (with no error code whatsoever). I'm guessing this is related to another issue I am seeing: the transfer speed will decrease linearly:
afbeelding

The official client suffers of no such issue, it seems.

PR #19, opened 2018-08-07
fix(socket): prevent connection from dying when the main thread is blocked

Lange:
It moves the UDP socket so a child process.

Whenever the main thread is blocked, our socket is unable to send messages or process received messages. This is a problem, since the ATEM will close or re-initialize our connection if we fail to respond to pings for about a second.

This PR moves the socket to a child process, so that the main thread can be blocked without our socket missing pings, therefore keeping our connection to the ATEM alive regardless of how hard the main thread is churning.

### DIFF
--- a/src/lib/atemSocket.ts
+++ b/src/lib/atemSocket.ts
@@ -95,7 +95,7 @@ export class AtemSocket extends EventEmitter {
 			this._socketProcess = null
 		}
 
-		this._socketProcess = fork(path.resolve(__dirname, 'atemSocketChild.js'), [], {silent: true})
+		this._socketProcess = fork(path.resolve(__dirname, '../socket-child.js'), [], {silent: true})
 		this._socketProcess.on('message', this._receiveSubprocessMessage.bind(this))
 		this._socketProcess.on('error', error => {
 			this.emit('error', error)
@@ -123,7 +123,7 @@ export class AtemSocket extends EventEmitter {
 			throw new Error('Socket process process does not exist')
 		}
 
-		return Util.sendIPCMessage(this, '_socketProcess', message)
+		return Util.sendIPCMessage(this, '_socketProcess', message, this.log)
 	}
 
 	private _receiveSubprocessMessage (message: any) {
--- a/src/lib/atemSocket.ts
+++ b/src/lib/atemSocket.ts
@@ -111,7 +111,7 @@ export class AtemSocket extends EventEmitter {
 			this._socketProcess = null
 		}
 
-		this._socketProcess = fork(path.resolve('dist/lib/atemSocketChild.js'), [], {silent: true})
+		this._socketProcess = fork(path.resolve(__dirname, 'atemSocketChild.js'), [], {silent: true})
 		this._socketProcess.on('message', this._receiveMessage.bind(this))
 		this._socketProcess.on('error', error => {
 			this.emit('error', error)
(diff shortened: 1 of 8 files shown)


Lange:
This is ready for other sets of eyes to review it and provide feedback/direction. I requested reviews from a few folks to get that process started.

src/lib/atemSocket.ts
```diff
@@ -34,29 +21,17 @@ export class AtemSocket extends EventEmitter {
 		this._debug = options.debug || false
 		this.log = options.log || this.log
 
-		this._createSocket()
+		this._createSocketProcess()
+
+		process.on('exit', () => {
+			if (this._socketProcess) {
+				this._socketProcess.removeAllListeners()
```
mint-dewit: We need to recreate the Socket process, possibly emit errors as well. And I guess resend commands that have not been acknowledged yet.

src/lib/atemUtil.ts
```diff
@@ -24,6 +26,22 @@ export namespace Util {
 		return value
 	}
 
+	export function subprocessSendPromise (subprocess: ChildProcess | null, message: any) {
+		return new Promise((resolve, reject) => {
+			if (!subprocess) {
+				return resolve()
+			}
+
+			subprocess.send(message, (error: Error) => {
```
mint-dewit: The promise resolval here is based on not receiving any errors. I don't think it's good enough to resolve for example a `Atem.connect()` call based on the fact that the child process _appears_ to have seen the request. I think we need to send a message id along, store the promise and resolve it when a succes message comes back from the worker thread. (Or perhaps use some library that already implements this)

src/lib/atemSocket.ts
Lange: This particular line is about when the _parent_ process exits. When our parent begins exiting, we need to remove our error handlers from the child. If we don't, our error handlers will throw an error every time we exit the program.

You're looking for https://github.com/nrkno/tv-automation-atem-connection/pull/19/files/3ce0992e68a4c03ef4cf1bcac2c52cba86d95417#diff-e7a164ad6407add743a708771e7eecb8R97

src/lib/atemUtil.ts
Lange: Agreed, I can work o nthat.

src/lib/atemSocket.ts
mint-dewit: Whoopsie, you're right. All these sockets and connections everywhere ;)
We should still consider if we have to resend any commands, I guess.

Lange: Yeah a resend mechanism would probably be good. I'll evaluate how complex it will be to implement.

Lange:
I just realized this also breaks the `_sentQueue` in `atem.ts`. It gets broken because `_localPacketId` in `atemSocket.ts` is now always `1`, and it never changes.

Lange:
I pushed a pretty substantial commit just now. It tries harder to re-send IPC messages if they fail, and it fixes the promise resolution of `_sendCommand`.

Lange:
Depends on https://github.com/DefinitelyTyped/DefinitelyTyped/pull/28042

Lange:
I've currently taken this as far as I can without further feedback.

The acknowledgement and promise resolution of commands has been restored (it was broken in the first few commits to this PR). However, the promises `connect` and `disconnect` commands on `AtemSocket` still resolve as soon as the IPC message has been successfully sent, as opposed to when we know the action has been fully carried out. This is because these two methods don't send `AbstractCommand` instances, which is what our current promise resolution system supports.

To implement better promises for `connect` and `disconnect`, one of these two things would need to be implemented:
1) refactor these methods to use `AbstractCommand` instances
2) implement a second, parallel system for tracking acknowledgement of just these two commands

Lange:
Also, this PR should be squashed when it is merged, so that `master` doesn't get cluttered up with my `wip` commits:
image

PR #22, opened 2018-08-16
feat: implement MacroRunStatus command (MRPr)

Lange:
It implements the MacroRunStatus command (`MRPr`).

This command is not yet implemented in the current release.

The state of a given macro can now be tracked. Specifically, these pieces of state:
```ts
export interface MacroPlayerState {
	isRunning: boolean
	isWaiting: boolean
	loop: boolean
	index: number
}
```

* **Other information**:
The primary use of this command is for keeping track of when a macro has begun and/or finished its execution. Useful for writing external code which reacts to macros being invoked.

### DIFF
--- a/src/commands/Macro/MacroRunStatusCommand.ts
+++ b/src/commands/Macro/MacroRunStatusCommand.ts
@@ -1,0 +1,27 @@
+import AbstractCommand from '../AbstractCommand'
+import { AtemState } from '../../state'
+
+export class MacroRunStatusCommand extends AbstractCommand {
+	rawName = 'MRPr'
+	index: number
+
+	properties: {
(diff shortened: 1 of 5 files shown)


Julusian:
I don't think this should be an array of state.
As far as I'm aware, only one macro can be running at a time, so all of tho fields in the command form a macro player status.

I could be wrong though, this is working from memory from a year ago when I last looked into this

Lange:
@Julusian That makes sense and explains why `index` is the last parameter in the buffer, not the first. PR updated.

Julusian:
Looks good to me

Lange:
Should I merge this directly to `master`, or rebase onto `develop` and merge there?

PR #26, opened 2018-10-10
feat: Use device ModelId to determine how many psu to track in state

Julusian:
I can't find anything in TopologyCommand that could indicate the number of psu the device has, so have had to go with a mapping from ModelId.

state.info.power will be an array with length matching the number of psu, so library consumers can rely on this length as the count of psu.


mint-dewit:
> I can't find anything in TopologyCommand that could indicate the number of psu the device has, so have had to go with a mapping from ModelId.

I'm not super happy about this, but it's really not something we can fix currently.
I've added a small bit of documentation, and let's just hope that if someone runs into issues with newer models they can find out where to fix. (Assuming that we don't buy every model of atem to test 😛)

Issue #23, opened 2018-09-01
Consider using `get-port` instead of Math.random


Lange:
* **I'm submitting a ...**
[x] bug report
[ ] feature request
[ ] question about the decisions made in the repository
[ ] question about how to use this project

* **Summary**
The `_createSocket` method currently relies on `Math.random` to choose its port ([`src/lib/atemSocket.ts#L130`](https://github.com/nrkno/tv-automation-atem-connection/blob/4086ddb0bda6133afbcfd4b847cafca5a9dc9f84/src/lib/atemSocket.ts#L130)). This is not guaranteed to work, and could result in startup errors if it chooses a port already in use.

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. StackOverflow, personal fork, etc.)
A better solution would be to use a library such as [`get-port`](https://www.npmjs.com/package/get-port). However, this would make the `_createSocket` method become asynchronous.

baltedewit:
Yeah, the way we do it could be a problem statistically seen. I think we need to take the basic idea `get-port` uses and adapt it to test `udp` ports.

`_createSocket` is an internal function so we can change it without any worries.

Lange:
Good catch, I didn't notice that `get-port` was specifically checking TCP ports only.

PR #36, opened 2018-11-02
feat: retry to bind to different port when port in use

mint-dewit:
This should fix potential connection issues arising due to #23

### DIFF
--- a/src/lib/atemSocketChild.ts
+++ b/src/lib/atemSocketChild.ts
@@ -124,10 +124,18 @@ export class AtemSocketChild extends EventEmitter {
 		if (this._maxPacketID < this._localPacketId) this._localPacketId = 0
 	}
 
-	private _createSocket () {
+	private _createSocket (): Promise<void> {
 		this._socket = createSocket('udp4')
-		this._socket.bind(1024 + Math.floor(Math.random() * 64511))
 		this._socket.on('message', (packet, rinfo) => this._receivePacket(packet, rinfo))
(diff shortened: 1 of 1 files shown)


haakonnessjoen:
I am probably stupid for asking. But can't you just remove the port argument from bind() I thought node tried to give you a random port automatically?

```Call socket.bind if you want to receive datagrams. socket.bind() will bind to the "all interfaces" address on a random port (it does the right thing for both udp4 and udp6 sockets). You can then retrieve the address and port with socket.address().address and socket.address().port.```

**Comment by mint-dewit** [comment]
I hadn't thought of it, to be honest. Do you happen to know if that will always give a free port too?

**Comment by haakonnessjoen** [comment]
Old documentation saying "it does the right thing", and new documentation not mentioning random problems with binding to taken ports sure hints to it though.

Also, in the source of dgram in Node, if you try to send data without binding first, it automatically does: 

```
if (state.bindState === BIND_STATE_UNBOUND)
    this.bind({ port: 0, exclusive: true }, null);
```

And I think they/we would see a lot of issue requests if people would randomly get "port already in use" errors when just sending udp without binding, which a lot of people do.

So I would believe strongly that this part of the code could be safely replaced with a bind().

mint-dewit:
Closing in favour of https://github.com/nrkno/tv-automation-atem-connection/commit/b7be165ffd452ef05df0c1ea931322e6649cf6b1

PR #39, opened 2018-11-26
Release next

### LINKED ISSUE #23: Consider using `get-port` instead of Math.random

### LINKED ISSUE #37: 100% CPU usage because of invalid parameter
https://github.com/nrkno/tv-automation-atem-connection/blob/f35250d6eb46287d3295c27a772b4a8c17c09181/src/lib/atemSocketChild.ts#L57

Because of an invalid parameter to Nanotimer, the retransmit timer is called on *every tick*. This makes the spawed client process utilize 100% cpu on a core.

The reason is that the interval argument should have been '10m' not '10ms'.

### LINKED ISSUE #38: Timers are not always stopped on disconnect()
https://github.com/nrkno/tv-automation-atem-connection/blob/f35250d6eb46287d3295c27a772b4a8c17c09181/src/lib/atemSocketChild.ts#L74

(In disconnect) The timers should be stopped even if the connection has not been established. The clearing of timers should be moved to above the "```if (this._connectionState === ConnectionState.Established)```" check. ( or in the else if you want :P )

If I run disconnect() on the instance that has not connected, because I am going to delete the instance, there will be freerunning timers that I have no way to stop.

mint-dewit:
Fixes #23 
Fixes #38 
Fixes #37

### DIFF
--- a/src/lib/atemSocketChild.ts
+++ b/src/lib/atemSocketChild.ts
@@ -72,19 +72,19 @@ export class AtemSocketChild extends EventEmitter {
 		return new Promise((resolve) => {
 			if (this._connectionState === ConnectionState.Established) {
 				this._socket.close(() => {
-					clearInterval(this._retransmitTimer)
-					clearInterval(this._reconnectTimer as NodeJS.Timer)
-					this._reconnectTimer = undefined
-
-					this._connectionState = ConnectionState.Closed
-					this._createSocket()
-					this.emit(IPCMessageType.Disconnect)
-
 					resolve()
 				})
 			} else {
 				resolve()
 			}
+		}).then(() => {
+			clearInterval(this._retransmitTimer)
+			clearInterval(this._reconnectTimer as NodeJS.Timer)
+			this._reconnectTimer = undefined
+
+			this._connectionState = ConnectionState.Closed
+			this._createSocket()
+			this.emit(IPCMessageType.Disconnect)
 		})
 	}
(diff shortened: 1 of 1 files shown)


haakonnessjoen:
Why did you remove the Nanotimer? It was just a wrong parameter?

mint-dewit:
It's primary use is for the SetTimeout in the ack'ing strategy, that has not been removed.

We don't require the resending of actual commands to be as accurate as the ack'ing strategy, since we only resend after 30ms anyway it is fine if the Interval takes some ms extra. In my testing the cpu usage was still at 100% when not sending any commands, even with the correct parameter.

PR #46, opened 2019-03-22
fixed super source box parameters (Crop Right can be 32000, not 3200)

Kohbrax:
Bug Fix

The library keeps throwing exceptions because of invalid super source range

It's documented in the unofficial ATEM-Protocol documentation: https://www.skaarhoj.com/fileadmin/BMDPROTOCOL.html under 'SSBP'. The range can be von 0 to 32000


Julusian:
Thanks, good spot :)

Kohbrax:
thanks, that was fast! :)

PR #48, opened 2019-04-01
Bugfix in audio mixer + added Master Channel

nytamin:
Bug fix: Handling of audio channels data
Feature: Add support for Master Channel

Breaking change: I renamed the `setAudioChannelXX` methods to `setAudioMixerInputXXX` so that they will better harmonize with the other audio commands (Input, Master, Monitor (not implemented yet) etc..)

### DIFF
--- a/src/atem.ts
+++ b/src/atem.ts
@@ -22,7 +22,7 @@ import { DownstreamKeyerGeneral, DownstreamKeyerMask } from './state/video/downs
 import * as DT from './dataTransfer'
 import { Util } from './lib/atemUtil'
 import * as Enums from './enums'
-import { AudioChannel } from './state/audio'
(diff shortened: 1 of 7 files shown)


nytamin:
Note: I never got the _followFadeToBlack_ to work (ref: https://github.com/LibAtem/LibAtem/blob/master/LibAtem/Commands/Audio/AudioMixerMasterSetCommand.cs).

But the master level work at least (tested on a Television Studio)

mint-dewit:
I've confirmed that this PR works with my Atem Television Studio HD. I had to push a small fix to the master command as that didn't set the flag on the master command, instead it always set it to 1, causing it to write gain every time instead of the appropriate properties. Now that this is fixed _fadeToBlack_ works as well.

PR #49, opened 2019-04-01
- fixed wrong number. borderHue can be a number between 0 and 3599.

Kohbrax:
fixed the wrong check of Border Hue inside the Protocol

The Applications crashes on wrong value

It uses now the correct value

* **Other information**:
see 'KeDV' in https://www.skaarhoj.com/fileadmin/BMDPROTOCOL.html


mint-dewit:
Whoops, I'm sure that git blame is on me ;)

Thanks for the PR! This will be in the next release.

PR #52, opened 2019-06-20
Tests for generated commands

Julusian:
This is a bit of a prototype currently, but it will want some discussions and this will be a good place for that.

The basic idea is to have a json file defining the commands in both json and serialized byte data. Then we either serialize the json, or deserialize the byte data depending on whether the command is one we send or receive.

This json file is being [generated](https://github.com/Julusian/atem-connection-test-generator/blob/master/AtemCommandTestGenerator/Program.cs) from my [C# library (LibAtem)](https://github.com/libatem/libatem) as that has [tests checking it against the official SDK](https://github.com/libatem/libatem.ComparisonTests) so should be correct.
Essentially, these tests verify that the command encoding matches LibAtem, but it doesn't directly guarantee they are correct, this is assuming that LibAtem ensures it has sufficient coverage against the SDK.

Some questions:
* With the json structure of commands varying slightly across the libs, which side should the conforming code be on?
* Should the json file only contain a whitelist of commands, that are implemented here?
* The json file currently has 10 random testcases for each command, is that a good idea?
* Perhaps it should be a file per command, otherwise the whole file will change if a single command needs regenerating

@baltedewit @Lange I expect you to both have some input on this


Julusian:
I'm not sure how to do the command serialization tests.
The rawName of the commands aren't exposed so it is possible to build a map of objects like is done for deserializing.

My thoughts are we could done one of
 * split each command into a serialize or deserialize. That would be breaking due to class name changes
 * expose the name used for serialization (I have tried this for [AuxSourceCommand](https://github.com/nrkno/tv-automation-atem-connection/blob/feat/tests/src/commands/AuxSourceCommand.ts) but it feels messy/confusing to expose a second name on each command
 * Something else?

mint-dewit:
> With the json structure of commands varying slightly across the libs, which side should the conforming code be on?

I'd like do to as little work as possible inside the tests, so it makes sense to optimize the json file. It is not exactly LibAtem's responsibility though, so we can make a small standalone conformer script too.

> Should the json file only contain a whitelist of commands, that are implemented here?

That doesn't matter too much. Ideally this library would contain all commands anyway. In the current situation it's just some extra lookups. On the other hand, not too hard to add if we go with the standalone conformer script.

> The json file currently has 10 random testcases for each command, is that a good idea?

Sure, in case we have some mistakes regarding 8bit vs 16bit integers this might be nice.

> Perhaps it should be a file per command, otherwise the whole file will change if a single command needs regenerating

Theoretically, yes. But would we regenerate that often and is that worth the extra bloat of the files?

> The rawName of the commands aren't exposed so it is possible to build a map of objects like is done for deserializing.

Seperating out the commands wouldn't have to be a breaking change if you consider that the deserializing commands are technically part of the public API but they are only used internally. They are part of the public serializable commands purely for organizational purposes. Keeping current command names for serializable commands and new ones for deserialization would require no changes from library users.

Julusian:
> > With the json structure of commands varying slightly across the libs, which side should the conforming code be on?
>
> It is not exactly LibAtem's responsibility though, so we can make a small standalone conformer script too.

It can go in the [generator program](https://github.com/Julusian/atem-connection-test-generator/blob/master/AtemCommandTestGenerator/Program.cs) then.
Although, I dont think all of it can be moved as the tests will need to know which id fields to compare as they sit outside of the properties object.

> > Should the json file only contain a whitelist of commands, that are implemented here?
>
> That doesn't matter too much. Ideally this library would contain all commands anyway. In the current situation it's just some extra lookups. On the other hand, not too hard to add if we go with the standalone conformer script.

Having them all creates 'noise' in the test results, as it produces a testcase that doesnt do anything, which I am not that keen on. But not having them will make it a lot harder to add new commands with more chance of tests being forgotten/missed. Perhaps I can do a better job at filtering out the unsupported ones so they dont show up

> > Perhaps it should be a file per command, otherwise the whole file will change if a single command needs regenerating
>
> Theoretically, yes. But would we regenerate that often and is that worth the extra bloat of the files?

Only when a bug is found in a command or something changed in one.

> > The rawName of the commands aren't exposed so it is possible to build a map of objects like is done for deserializing.
>
> Seperating out the commands wouldn't have to be a breaking change if you consider that the deserializing commands are technically part of the public API but they are only used internally. They are part of the public serializable commands purely for organizational purposes. Keeping current command names for serializable commands and new ones for deserialization would require no changes from library users.

Can the user not subscribe to individual commands being received? If not then that makes this easy.

Julusian:
Another question is where should the generator code be hosted. Should it stay as its own repo?

mint-dewit:
> Can the user not subscribe to individual commands being received? If not then that makes this easy.

The command is passed in the `stateChanged` event, so if a library user is using the class name in there that would be a breaking change, yeah

The other thing we can do is create a simple external mapping, but it would need to be maintained when commands are added.

Julusian:
ah, with the command to `stateChanged` then I dont think we can avoid this being breaking without making a bigger mess.

I'm not against a breaking change, but wanted to flag that before embarking on restructuring almost every command in case we wanted to avoid that.

Latest commit prototypes that on the Aux command, and starts to test a bunch of the serializable commands

krocheck:
> The command is passed in the `stateChanged` event, so if a library user is using the class name in there that would be a breaking change, yeah

This is the case for Companion but its a select-case, so it would be quick to update that to new names.

Julusian:
@baltedewit I am finding there are a few different ways we are building the buffers in serialize methods. Any objection to standarising them to look like https://github.com/nrkno/tv-automation-atem-connection/blob/master/src/commands/Inputs/InputPropertiesCommand.ts#L57 ?

Im also thinking of moving the prefixing with the command name out of those methods, now that the name is exposed

mint-dewit:
> Any objection to standarising them

Not at all, that is the way I build all new commands too. Just haven't gotten round to converting the rest.

> I'm also thinking of moving the prefixing with the command name out of those methods

I don't see any problems with this at this point.

Julusian:
@baltedewit I think this is ready for a proper review now. There are some test 'failures' still, so I have disabled testing those commands for now and shall fix them up later on.

I havent moved the json mutation logic to the generator yet, as it cant all be moved. Some of it is to know what the id properties are, as they live outside of the properties object. Potentially this array could also be improved a bit if we allow sharing the converters between commands, but that can get very confusing if not done well. The ids might be simple enough to do with a single lookup table though (and if not perhaps they should be conformed to be more consistent)

The only remaining question is Where should the generator project be located?

Lange:
Hi, sorry I haven't participated in this discussion yet. I've been very busy and making some life changes recently.

Things are beginning to settle down a bit, so let me know if my input is still requested. If so, a brief summary of the current status of things and what choices are left to be made would be very helpful!

mint-dewit:
> Where should the generator project be located?

If we keep the mappings for `atem-connection` inside the test files then it might make more sense to keep it as a LibAtem library so anyone wanting to use it for another language can do so.

There's a also a few more commands that are still failing:

- [x] **_pin**: discrepancy between atem-connection and LibAtem in the enum
- [ ] **AMMO**: Rounding error
- [ ] **KKFP**: LibAtem incorrectly(?) uses doubles instead of integers for typings
- [ ] **CAMI**: Rounding error
- [ ] **AMIP**: Rounding error
- [ ] **FTSU**: Unkown props getting overwritten by generator instead of ignored
- [ ] **TDdP**: Range validation errors (250 should be max, but 251 is generated)

Julusian:
@Lange Not a problem, both @baltedewit and I have been off for a few weeks, so nothing has been blocked. I think the versioning PR is more likely to warrant some input if you want to join in, this one has some refactoring but its pretty straightforward so I am happy with its current state.

> @baltedewit it might make more sense to keep it as a LibAtem library

Sure. Lets go with that for now, we can always move it later if we find it is becoming too atem-connection specific, or want to move the mappings out of the test files in here.

@baltedewit Be careful with fixing those failing tests, I dont remember exactly what the issues were, but I think I left them as they needed some testing to verify which side was at fault.

PR #53, opened 2019-06-25
Multiple version support

Julusian:
Moving discussions to a PR. @baltedewit @krocheck 

I've had a play around with what I was suggesting for how this should work. I havent tested anything yet, but this shows what I am thinking.

I haven't verified that the multiview changes will fit in nicely yet, but hopefully they will too. This is only a small subset of @krocheck's changes.

There is now an enum of versions, which makes comparisons and checking for certain versions very easy. Thinking about it, I don't see the value in the major and minor numbers we reported before, I think keeping it to the release versions is easier to understand. (Feel free to disagree on removing them)

Then as a user, whenever the ssrc state changes there is minimal difference between how v8 and v7.2 will behave on that. The boxes don't need separate commands as its a simple offset by another byte.
The properties and border has been split into a 7.x command, and the 2 for 8.0. The 7.x command then implements a new method `convertToLatestVersion` so that the stateChange event can emit with only the 8.0 version of the commands, simplifying the handling for the user.

On the sending side, the `src/atem.ts` methods are doing a version check so both versions will be used the same. But if manually creating the command, then the user needs to make sure to use the correct version. To me this is acceptable as this is a more advanced (and I expect less common) use case.

I'm not sure on the correct naming of the commands. This is going to be a large breaking change so I'm not worried about renaming this time, but when this happens in future it would be nice to not have to release a breaking change again. In theory this structure might allow us to keep backwards compatible versions for most changes, but it does depend on what Blackmagic change each version.

Any thoughts on this approach?


krocheck:
I have to re-state this.  Class names using 7X, 8X, etc. is going to be a convoluted mess in the long-run.  The reality is that every change has a real version number associated with it.  There were how many '7X' versions?  Blanket creating an Atem7XState and some of the associated classes makes no sense as a generic 7X state truly isn't a thing, nor will 8X be a thing in the long run.

In that respect, the 2.28, etc. numbering for the API makes a lot of sense because the changes between v7.5.2 and v8 with respect to the API are not major when looking at the API as a whole.  We would expect a v3.00 to include drastic breaking changes.  The public software versions account for new switchers and some changes as a result.  Granted, that is the versioning the public addresses, so internally cross-tabbing the API to software version is necessary.  But its worth remembering that we're API users, so designing around that makes the most sense to me.

This is why I circle back to the Command classes handling this internally, and I know this is somewhat built into this approach.  If the Command classes internally want to initialize their own 7X and 8X state-like classes to deal with hard discrepancies then that works.  Internal is the key, though.

In Companion, I don't plan on wanting to do a bunch of version detection and deal with different calls based on switcher firmware version.  Granted we can for certain breaking things like SSrc, but I'm not going to filter down capabilities based on firmware version.  We already do enough of that based on model detection.  Cross-tabbing model and firmware is just going to bloat certain things unnecessarily.

Julusian:
I have been thinking about this, and I partially agree with you.
Versioning the state will give a lot of library users more work to handle that, so is something we should really try to avoid doing. I expect we should be able to stick to that pretty easily, but that depends on changes in the protocol so cant guarantee anything.

As for commands, I need to think a bit more on that..

As a related thought, I am thinking that the stateChanged event should be changed a bit to not pass the command causing a change. Something simpler such as the path within the state of what changed will give the user the same information, without needing any knowledge about any of the commands. For example, instead of SuperSourceBoxParametersCommand it would pass `video.supersource.0.box.0` (or maybe `['video', 'supersource', 0, 'box', 0]`). Then the user can still see what has changed, but without seeing a command.

In companion you are using the normal api, and whichever way we go that will be consistent for both versions.

krocheck:
I don't disagree with using path-based information for stateChanged.  At that point you're basically doing what [atemOSC](https://github.com/danielbuechele/atemOSC) does.  That certainly would change the parsing in Companion from a select-case over to if-else with a bunch of string matching.  I would recommend adding that as a touch point in the return and let the library user decide to use one method over the other.  I think it should be a string like `video.supersource.0.box.0` as you can do a quick split if an array is needed.  Long-term I definitely would use the path-based addressing.

Julusian:
Yes, either of the proposed changes to stateChanged will be a little harder to match than the current method, but it is a single regex comparison. It is also a lot more descriptive and makes the parameter more easily predictable.
Converting in either direction between a string and array is trivial, but I think a string is easier to match in a single if statement, so is a better choice. 

Back to command versioning. With this stateChanged event, there is very little reason for a library user to ever need to think about the underlying commands. Most users will be fine with the exposed api. The main reason I can think where they might be used is to be able to batch or queue sending. I don't have a solution for batching right now, but queueing can be achieved around the exposed api too.
The only other use case I can think of is to write something like [this](https://github.com/nrkno/tv-automation-atem-state). That library will be a perfect example of what it will be like to support multiple versions of commands, so perhaps the best approach here is for me to try it there and see how it feels to work with.

I very much do not want the commands to have to parse from multiple atem commands, the api of those commands does not work well for that because the typings of the properties object will cause a lot of issues. If that is what we want, then the received commands should be replaced with 'parsers' which deserialize and apply to the state in one step. But this could end up with some large amount of special case handling for different versions, so could easily become a mess. And we need to be careful to not break older versions when fixing/changing newer versions, so the less interlinked the code is the less likely we are to break something. The unit tests I have been writing will help, but they dont cover many versions so wont catch everything. I am open to this idea though, it would be worth prototyping.

Outputting multiple atem commands doesnt feel right either. That feels like the target for another layer of abstraction, something like that offered by atem-state library. So I feel like we should call direct command manipulation an internal api, and suggest atem-state as something to use instead.

But I dont think that multiple versions of commands in the current structure is a large maintenance burden. The only additional cost I can see is that their applyToState methods will need occasional updating, but beyond that they will sit in existence until we drop support for that version and delete them.

If you think I have overlooked something please say so, or if you can think of some use cases I have not thought of.

PR #55, opened 2019-09-17
WIP: serializing commands without update props

Julusian:
@baltedewit Something for one of us to look into sometime.
This would be a better fix for `autoDownstreamKey()` than 71c61f66529168615e99622da32ebc14e17c4918


Julusian:
Closing in favor of #60

PR #57, opened 2019-09-18
feat: add listVisibleInputs method

### LINKED ISSUE #54: Consider adding API for enumerating all visible Inputs in PVW and/or PGM
* **I'm submitting a ...**
[ ] bug report
[x] feature request (I'll code it and open a PR if we think its a good idea)

* **Summary**
A common task is writing code which reacts to which Inputs are present in PGM and/or PVW. This is harder than it seems at first glance, because we can't simply use the `previewInput`/`programInput` property of the ME state object, because that wouldn't account for keyers and SuperSources. To figure out what Inputs are actually displaing in PVW/PGM right now, we have to loop through all keyers and SuperSources.

This is a common enough task that I think it should be provided by this library. This is my first thought at what this API would look like:

```ts
atem.on('stateChanged', state => {
    const me = state.video.getMe(0);

    // This includes the following Inputs:
    // 1. The current `programInput`
    // 2. If `programInput` is a SuperSource, then it also includes all inputs being used by that SS (recursively)
    // 3. Any inputs currently being shown as DSKs or USKs
    console.log(me.getVisibleInputs('program')); // => logs an array of numbers

    // Same as above but for PVW instead of PGM.
    console.log(me.getVisibleInputs('preview')); // => logs an array of numbers
});
```

I'm not completely married to this API design, and am very open to feedback and suggestions.

cc @baltedewit and @Julusian for their thoughts

Lange:
Feature

#54

* **Other information**:
Closes #54 

To test this out locally, create and run a script like the following:

```js
'use strict';

const {Atem} = require('./');
const myAtem = new Atem();
myAtem.connect('192.168.1.52');

let last = [];
myAtem.on('stateChanged', () => {
	const freshPgm = myAtem.listVisibleInputs('program', 0);
	const freshPvw = myAtem.listVisibleInputs('preview', 0);
	const concatted = freshPgm.concat(freshPvw);
	if (JSON.stringify(last) !== JSON.stringify(concatted)) {
		console.log()
		console.log('pgm:', freshPgm)
		console.log('pvw:', freshPvw)
		last = concatted;
	}
});
```


src/atem.ts
```diff
@@ -526,6 +526,52 @@ export class Atem extends EventEmitter {
...
+			lastSize = inputs.size
+			inputs.forEach(inputId => {
+				if (!this.state.inputs[inputId]) {
+					// Data isn't hydrated yet, we'll get 'em next time.
+					return
+				}
+				const portType = this.state.inputs[inputId].internalPortType
+				switch (portType) {
+					case Enums.InternalPortType.SuperSource:
+						const ssrcId = 0 // TODO: don't hardcode this
```
Lange: Need to figure out an algorithm here like the one being used in the `case Enums.InternalPortType.MEOutput` block below. To figure out this algorithm, I need to know what the input ID of the second SuperSource on the new big boy ATEM is.

src/atem.ts
```diff
@@ -554,4 +600,85 @@ export class Atem extends EventEmitter {
...
+
+			// This is the only USK type that actually uses the cutSource.
+			if (usk.mixEffectKeyType === Enums.MixEffectKeyType.Luma) {
+				inputs.add(usk.cutSource)
+			}
+		})
+
+		// DSKs only show up on ME 1,
+		// so we only add them if that's the ME we are currently processing.
+		// TODO: Is this still true for the new Constellation ATEMs?
```
Lange: Someone who knows more about the new ATEMs needs to fact-check this.

src/atem.ts
```diff
@@ -554,4 +600,85 @@ export class Atem extends EventEmitter {
...
+		inputs.add(meRef[mode === 'program' ? 'programInput' : 'previewInput'])
+
+		// Upstream Keyers
+		Object.values(meRef.upstreamKeyers).filter(usk => {
+			if (mode === 'program') {
+				return usk.onAir
+			}
+
+			// Pretty gross bitwise operations in this next line, be warned.
+			// TODO(Lange): This feels fragile but I'm unsure how to make it more robust.
```
Lange: Is this fine? Am I worrying too much?

src/atem.ts
```diff
@@ -554,4 +605,91 @@ export class Atem extends EventEmitter {
...
+			}).forEach(dsk => {
+				inputs.add(dsk.sources.fillSource)
+				inputs.add(dsk.sources.cutSource)
+			})
+		}
+
+		// Compute what sources are currently participating in a transition.
+		// We only care about this for PGM.
+		if (meRef.inTransition && mode === 'program') {
+			// The previewInput is always participating in the transition.
```
Lange: Just realized this is not true and that I need to read `transitionProperties.selection` to determine when to set this.

Lange: Should be fixed now.

src/atem.ts
Julusian: I believe the ids are 6000 and 6001

src/atem.ts
Julusian: Yes, I believe this is true. I can't see the output, but they all show up only on the ME1 control page.

src/atem.ts
```diff
@@ -554,4 +605,104 @@ export class Atem extends EventEmitter {
...
+			}
+		} else {
+			inputs.add(meRef.programInput)
+		}
+
+		// Upstream Keyers
+		Object.values(meRef.upstreamKeyers).filter(usk => {
+			// Pretty gross bitwise operations in this next line, be warned.
+			// TODO(Lange): This feels fragile but I'm unsure how to make it more robust.
+			const isPartOfTransition = meRef.transitionProperties.selection & (2 ** (usk.upstreamKeyerId + 1))
```
Julusian: I dont think this can be changed much, bitwise operations is just how something like this has to be done. 
I noramlly see this form though, so to me this is immediately more recognisable `& (1 << (usk.upstreamKeyerId + 1))`. Not sure if it is overall better though.

Other than that perhaps just pull that part into a variable to explain it a little better:
```
const keyerMask = 1 << (usk.upstreamKeyerId + 1)
const isPartOfTransition = meRef.transitionProperties.selection & keyerMask
```

src/atem.ts
```diff
@@ -554,4 +605,104 @@ export class Atem extends EventEmitter {
...
+		// TODO: Is this still true for the new Constellation ATEMs?
+		if (meId === 0) {
+			Object.values(this.state.video.downstreamKeyers).filter(dsk => {
+				if (mode === 'program') {
+					return dsk.onAir || dsk.inTransition
+				}
+
+				if (!dsk.properties) {
+					// Data isn't hydrated yet, we'll get 'em next time.
+					return
```
Julusian: Should this return false, or is everything happy with undefined?

src/atem.ts
```diff
@@ -554,4 +605,104 @@ export class Atem extends EventEmitter {
...
+						inputs.add(meRef.transitionSettings.DVE.keySource)
+					}
+					break
+				case Enums.TransitionStyle.WIPE:
+					if (meRef.transitionSettings.wipe.borderWidth > 0) {
+						inputs.add(meRef.transitionSettings.wipe.borderInput)
+					}
+					break
+				case Enums.TransitionStyle.STING:
+					inputs.add(meRef.transitionSettings.stinger.source)
```
Julusian: These ids need translating, as the `source` is just an index of the media player https://github.com/LibAtem/LibAtem/blob/master/LibAtem/Common/StingerSource.cs

Julusian:
Ive read through the code and added a few comments.

Could you also have a look into preview transition. I believe that changes what the pvw output shows, so it would be good to check that is covered.

Also, I am a little concerned in general about the length of this atem.ts file. I think it would be better to have this code in a seperate file to help keep the size down, but then we can wrap it so that it is exposed on the Atem class

Lange:
@Julusian Thanks for the review! I agree with all your feedback. I'm hitting the road in a few days for a show, but when I get back I should be able to finish this up.

Lange:
@Julusian All feedback addressed, LMK what you think of the changes.

Lange:
Here's one of my current use cases for this feature: https://youtu.be/TAXwDG5BJuo

I've implemented audio-follow-video in a setup consisting of an ATEM and an X32, which of course do not integrate at all and can't normally do this.

Julusian:
I have an idea for some tests that I am playing with.
It appears I forgot to say before, but the atem sends a command which reports basically this data in a less useful fashion as it can only report the me1 version of this. The benefit of this method is that it can be run on just me2, which will be useful.

Julusian:
I've added some test cases here.
A couple of the cases are failing currently, I didn't try looking into what was wrong as I wasnt sure which is correct.

What do you think of tests in this style? Only testing for ME1 currently, doing ME2+ will be a much more manual process.
Generating the cases is really easy, as they are dumped directly from the atem.

Lange:
This style seems good to me -- and even if they aren't 100% ideal, the fact that you say they are easy to generate has value to me, as this could be SUPER tedious and I don't think either of us want that.

Issue #40, opened 2018-12-03
Socket-child can be left orphaned on parent crash


Julusian:
* **I'm submitting a ...**
[*] bug report

* **Summary**
I appear to have had another possible occurrence of #30. I noticed my cpu fan running pretty high and turns out I have 2 atem-connections still open and maxing out a core each.

* **Other information**
Running v0.8.0
I had playout-gateway crash a couple of times on me earlier on today, this could have been caused by that, but the stack trace is unrelated

I found I could attach a debugger to the process (https://hackernoon.com/debugging-node-without-restarting-processes-bd5d5c98f200), and the console is being flooded with:
```
warning.js:18 (node:8448) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'emit' of undefined
    at log (/home/julus/Projects/tv-automation-playout-gateway/node_modules/atem-connection/dist/lib/atemSocketChild.js:80:14)
    at Object.onFailedAttempt (/home/julus/Projects/tv-automation-playout-gateway/node_modules/atem-connection/dist/lib/atemUtil.js:74:21)
    at Promise.resolve.then.then.error (/home/julus/Projects/tv-automation-playout-gateway/node_modules/p-retry/index.js:44:14)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)

warning.js:18 (node:8448) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 30132)
```
And a cpu profile: [CPU-20181203T211452.zip](https://github.com/nrkno/tv-automation-atem-connection/files/2641308/CPU-20181203T211452.zip)
I cant think of anything else I can look at to get more info

Julusian:
I havent seen this in a while, so I assume those fixes worked

Issue #30, opened 2018-10-19
Socket-child can be left orphaned on parent crash


Julusian:
* **I'm submitting a ...**
[*] bug report

* **Summary**
I noticed today that I had a few socket-child processes running on my system with no node processes running that would be utilising an atem-connection.
![image](https://user-images.githubusercontent.com/1327476/47246836-ef4d3f00-d3f7-11e8-9b2c-732644d9e53a.png)
I have tried to reproduce this and have managed to do so if the parent application crashes with an unhandled exception, but there could be other ways of achieving the same result.

I'm not sure how the memory usage got so high. I suspect that they have been running for a few days and have a huge number of commands to send to the parent buffered.

mint-dewit:
Tested in Ubuntu 18.04 and fixed in `atem-connection` 0.7.0

Please reopen if the issue occurs again!

Issue #15, opened 2018-08-06
incorrect parsing of ATEM data


nytamin:
* **I'm submitting a ...**
[X] bug report
[ ] feature request
[ ] question about the decisions made in the repository
[ ] question about how to use this project

* **Summary**
Some strange data sent from my ATEM Television Studio causes the state to go bananas.

* **Other information**
Among others, this data is sent into MixEffectKeyFlyPropertiesGetCommand.deserialize():
Buffer(0x509be28400000000)
this creates
	this.mixEffect = 80
which in turn when calling state.video.getMe(this.mixEffect)
	creates ME[80], effectively making ME into a sparse array, causing problems further down the line

I'm appending a [wireshark trace,](https://we.tl/Y5vO7Dkzh9) there are some other trange data in there as well, causing the state.video.ME[] to become [ [Object], <72 empty items>, [Object], <6 empty items>, [Object], <26 empty items>, [Object] ]
on the indexes [ '0', '73', '80', '107' ]

mint-dewit:
The current status on this is that we throw an error when we find incorrect data. On some models this can however be mildy annoying as there are many errors (Ex. #24)

The errors on the user side of the library do not mean the user can do anything against this, this is really an error on the atem side of the connection and we unfortunately are not in a position where we can improve that. For this reason I think we could change the error throwing behaviour to a warning emitter or a `parserError` emitter. This could be something used for debugging or even logging in production systems but would not be as annoying for library consumers working on models that emit a lot of errors.

Julusian:
A prerelease of v2.0.0 has just been pushed to npm under the nightly tag `yarn add atem-connection@nightly` that should resolve this, along with other problems.
The parsing has been made a lot less picky on values to avoid errors like this, and any out of range ids are now debug warnings rather than errors.

We are happy that v2.0.0 should be stable, but want to spend a bit more time giving it some realistic testing before doing a proper release of it.

PR #60, opened 2019-10-08
feat: strict tsconfig and use threadedClass

### LINKED ISSUE #24: Unable to connect to Atem
* **I'm submitting a ...**
[x] bug report
[ ] feature request

* **Summary**

I am trying to use atem-connection to connect to a Atem TVS using this code:

	const { Atem } = require('atem-connection');
	const myAtem = new Atem({ debug: true, externalLog: console.log });
	myAtem.connect('10.0.0.35');

If the mixer is off, node is logging 'reconnect' every second, 
but if the mixer is on node crashes with this error:

events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: Number too small: 0 < 1
    at Object.parseNumberBetween ( - - - \node_modules\atem-connection\dist\lib\atemUtil.js:23:19)
    at TransitionDVECommand.deserialize ( - - - \node_modules\atem-connection\dist\commands\MixEffects\Transition\TransitionDVECommand.js:17:32)
    at AtemSocket._parseCommand ( - - - \node_modules\atem-connection\dist\lib\atemSocket.js:162:21)
    at AtemSocket._parseCommand ( - - - \node_modules\atem-connection\dist\lib\atemSocket.js:171:18)
    at AtemSocket._parseCommand ( - - - \node_modules\atem-connection\dist\lib\atemSocket.js:171:18)
    at AtemSocket._parseCommand ( - - - \node_modules\atem-connection\dist\lib\atemSocket.js:171:18)
    at AtemSocket._parseCommand ( - - - \node_modules\atem-connection\dist\lib\atemSocket.js:171:18)
    at AtemSocket._receivePacket ( - - - \node_modules\atem-connection\dist\lib\atemSocket.js:130:18)
    at Socket._socket.on ( - - - \node_modules\atem-connection\dist\lib\atemSocket.js:110:60)
    at emitTwo (events.js:126:13)

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. StackOverflow, personal fork, etc.)

Julusian:
* **Remaining**
- [x] Parse/remove the newly added properties to the AtemCapabilites object.
- [x] How should we handle applyToState failing? That be ignored or logged in a way that doesnt trigger an error in cases that are expected (ie, no dve and jibberish data)
- [x] Use release version of threadedClass

Some non-blocking tasks moved to https://github.com/nrkno/tv-automation-atem-connection/issues/73

This enables some stricter typescript rules. Primarily to enable `strictPropertyInitialization` to ensure that all class properties are initialised before they can be used.
Additionally, fork has been replaced with threadedClass which is more durable and allows for using workers instead of fork when using recent enough node.

The connection handling logic has been thoroughly reviewed and many bugs have been fixed. There were some problems with retransmits when the packet id would wrap, which was causing instability with media upload, and other possible connection deadlocks if the connection is unstable and started to hit some timeouts.

Data parsing has been loosened, as we cannot guarantee data ranges, and that they wont change. We now parse the command, and then decide whether to apply it to the state based on whether the ids are within known good ranges for the current device. We assume that any non-id value received is valid, and we assume all values in outbound commands are within range (including ids). This should fix #24 #15 #61 and prevent similar issues cropping up in future.

Typescript is a bit loose in its checks, so some fields can be undefined even though their typings do not indicate that.

Using fork to open the child process has caused some stability problems, with it getting orphaned under certain circumstances. #30 #40 #68 #65 (possibly more)

* **Other information**:

This is a massive overhaul of a majority of the library.
Every command class had to be refactored to allow the typings to work. AbstractCommand has been replaced with a couple of separate base classes & interfaces. Different for each of serializable and deserializable commands.
Other benefits of this is that various properties are now readonly when possible, and there is less dependencies on ensuring 'random' properties are set at the right points.

Threadedclass does come with 4-5% more cpu usage (19% vs 14%) than the fork implementation while doing still media uploads. But with that we get a safer wrapper around the child process, both in typings and in durability. The performance drop hasnt been investigated fully yet, but that could be done if this is found to be a problem.

Additionally, media is now a lot quicker to upload than previously.
in node 8 with fork I was getting ~1400ms per frame, and that is down to ~1000ms when running this in node 12. node8 still performs around ~1400ms, the benefit here comes only when running newer node.


src/enums/index.ts
```diff
@@ -213,7 +213,8 @@ export enum TransferMode {
 	NoOp,
 	Write,
 	Clear,
-	WriteAudio = 256
+	WriteAudio = 256,
+	TEST = 0x0201
```
Julusian: remove this multiviewer value

Julusian:
@baltedewit I am wondering if AtemState and its children should be all interfaces, with the helper functions moved next to it.
This way we can allow/support the state to be cloned by whatever method desired, and not have to worry about the methods disappearing off it.

mint-dewit:
> How should we handle applyToState failing?

I think we should follow the idea that we should only error for a case that we did not expect to find. So in the case for a device not having a DVE that should be handled / ignored, possibly applyToState shouldn't even throw. In any other case it's unexpected and should be logged as an error.

> I am wondering if AtemState and its children should be all interfaces, this way we can allow/support the state to be cloned by whatever method desired without the methods disappearing off it

We could do that, but at the same time it's not a requirement for any of the current users. We could also wait until someone needs to do that. I imagine it can be done in a non-breaking way

Julusian:
> > How should we handle applyToState failing?
> 
> I think we should follow the idea that we should only error for a case that we did not expect to find. So in the case for a device not having a DVE that should be handled / ignored, possibly applyToState shouldn't even throw. In any other case it's unexpected and should be logged as an error.
> 

yeah, so should that be a warn/info/log/debug/trace level message or not at all logged? 
At the very least I need to create a new error type for it so that we can easily filter it out and not ignore real errors.

> > I am wondering if AtemState and its children should be all interfaces, this way we can allow/support the state to be cloned by whatever method desired without the methods disappearing off it
> 
> We could do that, but at the same time it's not a requirement for any of the current users. We could also wait until someone needs to do that. I imagine it can be done in a non-breaking way

I would like to throw in the unit-tests as a 'user' of this requirement.
There is currently a bunch of use of `class-transformer` to convert states stored as json files for tests into the real classes. 
https://github.com/TipoftheHats/atem-compositor is another example. In that the state is sent over IPC into the ui. I dont expect this to be an uncommon use case.

mint-dewit:
> yeah, so should DVE related failures be a warn/info/log/debug/trace level message or not at all logged?

It makes sense to add debug logging for it, I imagine a scenario where someone wants to make sure the handling of those failures is or is not the root cause of some bug they are investigating and in that case it would be nice to have that logged.

If there are current usecases for the serializing/deserializing of state object I guess it's a good idea to make that possible. I think we could either define the interfaces, have classes with initializers and move out the helper methods to be utility functions _or_ define everything as interfaces and have a utility function for generating a state with initialized values.

Julusian:
@baltedewit sorry for the large touching every command changes happening again.

This now has the state purely as interfaces, with AtemStateUtil containing all the `get*` methods we had scattered on the state before. I am not 100% sure this is the best right now, I shall update atem-state to this to check it feels ok from another library (it may not).

Then I have made a custom Error type for the expected invalid id errors thrown in applyToState. And changed how logs are passed out of the library. It now has 3 levels of emit for logs ('error', 'info' and 'debug')

Julusian:
Change pushed to atem-state to match this state refactor.
I think it is ok actually. It was pretty painless to convert to.

mint-dewit (approved):

mint-dewit:
Both changes look good to me!

PR #70, opened 2019-12-15
feat: add support for Time command

peschuster:
The Constellation 8K model has no connector for external time code, but since version 8 of the ATEM software, time code can be set from the software. Unfortunately, the software only allows resetting the time code to 0, but the device can handle all values set using the Time command.

This also works on other ATEM models (tested with TVS HD).


Julusian:
Thanks, this looks good.

State changed path should be `info.time`
I have seen this command received in older versions in the same packet as command responses, so it would be better to call it something like `lastTime` or `lastChangeTime` on the state to make it clear it is only updated periodically.
Ideally we should set minimum version on the command to v8, but as it is received in older versions, so I'm not sure how best to handle that. @baltedewit 

Based on the sdk api, I suspect that the remaining 4 bytes contains a 'drop-frame' bool. I cant remember if that was a set or get method, so it may only be valid in one direction.

peschuster:
I updated the path in the state change.

`GetTimeCode` in the SDK has a parameter `dropFrame` of type bool, but I am not sure how it is encoded in the four bytes. My observations:
- the last three 3 bytes stay constant until a power cycle
- the first unknown byte toggles between 0 and 1, although video mode is 1080i50 and should not have any drop-frames in time code(?)

I also added the "TiRq" command. The software executes this command periodically after the section "Timecode" is opened and the device responds with a "Time" packet.

mint-dewit:
I think the idea is to assume all new features are based on v8. So long as it doesn't absolutely break the library it doesn't _need_ a minimum version. At the same time, adding a minimum version wouldn't hurt anyone.

Julusian:
@baltedewit the challenge is that it is valid for serialization in 8.0+ but deserialization in older. It feels excessive to have 2 copies of the command to fix that, so I am tempted to just leave it as is.

@peschuster It looks like that toggling byte is a field marker, but its hard to be certain especially as the sdk doesnt appear to expose that.
The byte after that is the dropFrame bool.

Julusian:
I have squashed and merged this to develop.
I should be able to do a release later this week

Julusian:
@peschuster I just found that the software (in 8.1 at least) does allow setting the timecode to anything. Its not the most obvious, but select the box that is `00:00:00:00` and start typing. then press set

peschuster:
> @peschuster I just found that the software (in 8.1 at least) does allow setting the timecode to anything. Its not the most obvious, but select the box that is `00:00:00:00` and start typing. then press set

Oh, you're right. It is the same with 8.0.2. I tried to edit the field but did not realize I could just start typing. Thanks for the hint!

PR #111, opened 2021-04-18
feat: more commands

Julusian:

Feature

Implements various additional commands:
* Colour generators
* Advanced Chroma Keyer
* Multiviewers
* Classic audio
* Fairlight audio

* **Other information**:

This includes a couple of small state object changes, that are technically breaking.


Julusian:
@baltedewit Would you mind having a quick look at the state object, to see if there is anything you disagree with? 
Also, I marked a bunch of methods as deprecated in the atem class. As master is about to be v3, do you think we should take the opportunity to remove them, or should we hold off until v4 so that there is some warning of it?

src/__tests__/util.ts
```diff
@@ -62,6 +62,33 @@ export function createEmptyState(): AtemState {
...
+			equalizerBands: [undefined, undefined, undefined, undefined, undefined],
+			equalizerGain: 0,
+			equalizerEnabled: false,
+			makeUpGain: 0,
+			faderGain: 0,
+			followFadeToBlack: false,
+		},
+	}
+
+	for (let i = 0; i < 6000; i++) {
```
mint-dewit: 6000 empty object seems like a lot :)

src/state/fairlight.ts
```diff
@@ -7,37 +7,100 @@ import {
...
-	readonly equalizerBands: number
+	gain: number
+	balance: number
+	faderGain: number
+
 	equalizerEnabled: boolean
 	equalizerGain: number
+	readonly equalizerBands: Array<FairlightAudioEqualizerBandState | undefined>
+
 	makeUpGain: number
```
mint-dewit: Makes more sense to have it in `FairlightAudioDynamicsState` to me

src/state/fairlight.ts
```diff
@@ -7,37 +7,100 @@ import {
...
 	stereoSimulation: number
 
-	readonly equalizerBands: number
+	gain: number
+	balance: number
+	faderGain: number
+
 	equalizerEnabled: boolean
 	equalizerGain: number
+	readonly equalizerBands: Array<FairlightAudioEqualizerBandState | undefined>
```
mint-dewit: Might be nice to bundle all eq related properties in one object too?

mint-dewit (commented):
I'm fine with throwing out those properties in the first breaking release, the API hasn't changed that much after all

src/state/fairlight.ts
Julusian: I agree, but it makes the state updates from some commands a bit more complex, as the makeUpGain is part of the general source/master command.

Ill have a play to see how messy it becomes, it might not be as bad as I remember

src/__tests__/util.ts
Julusian: its not pretty, but was necessary to be at lease 2000 for the tests to be happy.
Otherwise the tests running applyToState would fail at https://github.com/nrkno/tv-automation-atem-connection/blob/feat/bitfocus-commands/src/commands/Fairlight/FairlightMixerSourceCompressorCommand.ts#L82.

It will probably be possible to decrease that number (I dont remember how high the ids can go), but for the cost of at most a couple mb of ram during the tests I dont see the benefit

Julusian:
@baltedewit thanks, Ill look at resolving those comments and shall strip out the deprecated methods

src/__tests__/util.ts
mint-dewit: I see this file is only used in testing, not as important as I thought then!

src/state/fairlight.ts
Julusian: yeah its a bit fiddly, but not too bad really.
In atem-connection it is ok, the state updates simply become more complex to spread the properties around.

It mostly comes down to handling it in atem-state, where it has to do the reverse before doing the diffing. But again, not too bad so im going to go with it so we have a cleaner state

src/__tests__/util.ts
Julusian: After fixing a bug, I discovered that I also need to create the correct sources inside these, with the index being some hard to guess string. So I changed this to do some light inspection of the command it is for instead

PR #121, opened 2021-11-15
feat: add support for Media Pool settings

alvancamp:
It adds support for the Media Pool settings, which is this part of the ATEM Software Control UI:
image

These commands (`CMPS` and `MPSp`) are unsupported.

These commands are supported.

* **Other information**:
At this time, tests are not passing. I'm unsure why. Help/guidance would be appreciated.

### DIFF
--- a/src/state/util.ts
+++ b/src/state/util.ts
@@ -40,7 +40,11 @@ export function Create(): AtemState {
 		},
 		settings: {
 			multiViewers: [],
-			videoMode: 0
+			videoMode: 0,
+			mediaPool: {
+				maxFrames: [0, 0, 0, 0],
+				unassignedFrames: 0
+			}
 		}
 	}
 }
--- a/src/state/util.ts
+++ b/src/state/util.ts
@@ -41,10 +41,7 @@ export function Create(): AtemState {
 		settings: {
 			multiViewers: [],
 			videoMode: 0,
-			mediaPool: {
-				maxFrames: [0, 0, 0, 0],
-				unassignedFrames: 0
-			}
+			mediaPool: undefined
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


src/state/util.ts
```diff
@@ -40,7 +40,11 @@ export function Create(): AtemState {
 		},
 		settings: {
 			multiViewers: [],
-			videoMode: 0
+			videoMode: 0,
+			mediaPool: {
```
Julusian: I would prefer the mediaPool object to be optional, so that it is undefined rather than filled with zeros when we havent received the command

Julusian (commented):
Looking good so far!

For the failing test, I think I remember they changed the MPSp and CMPS commands in 8.0 (because of the constellation). I expect if you add a line like this https://github.com/nrkno/tv-automation-atem-connection/blob/v2.x/src/commands/SuperSource/SuperSourcePropertiesCommand.ts#L78 to both commands, it should resolve it.

For reference, that test is trying to parse every command received from a 4me4k running 7.5.2 (and doing a bunch of other models and firmware combinations). The error indicates that the buffer received is 2 bytes, not the 8 you are expecting.
We have no interest in supporting anything older than the latest in new code. The line I linked will tell the parser to not try using this implementation on older firmware versions.

src/state/util.ts
alvancamp: Done!

alvancamp:
Thanks for the quick review and feedback! I've gone ahead and implemented the suggested changes. The tests are still failing, but in a different way:
image

Julusian:
> Thanks for the quick review and feedback! I've gone ahead and implemented the suggested changes. The tests are still failing, but in a different way: image

Ah yes. Those tests are some autogenerated tests against [my c# library](https://github.com/libatem/libatem) which was also missing the minimum version property. 
I've fixed the test data so it should be happy now

Issue #110, opened 2021-04-06
Clip upload lazy frame loading


Julusian:
* **I'm submitting a ...**
[ ] bug report
[x] feature request
[ ] question about the decisions made in the repository
[ ] question about how to use this project

* **Summary**
When uploading a clip to the atem, all the frames must be presented to the library at once. While this is optimal for upload speed, it will require very large amounts of memory when doing long clips (filling a constellation in 1080p will take 12gb of ram)

* **Other information**

https://github.com/nrkno/tv-automation-atem-connection/blob/abcff0c74a51b3ffb4880106b87109a7fc72b9ec/src/dataTransfer/dataTransferClip.ts#L12
Instead this should be able to take an `Array<Promise<DataTransferFrame>>`

There could be a strategy in place to preload _x_ frames to keep the uploader running smoothly.

What should happen if a frame promise rejects? Should the whole clip be invalidated and aborted?

Julusian:
Implemented with #122

PR #122, opened 2021-11-19
feat: add support for using generators with uploadClip

alvancamp:
Feature

`uploadClip` only accepts an array of RGBA framebuffers.

`uploadClip` now also accepts a (synchronous) [generator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*). Example usage:
```js
/*
Movie to tga sequence
ffmpeg -i inFileName.mov outFileName%03d.tga

Convert image-sequence (unix one-line bash)
for i in *.tga; do ffmpeg -i "$i" -pix_fmt rgba -f rawvideo "$(basename "$i" .tga)".rgba  ; done
*/
// Edit this to be a path on your computer that points to a directory of raw *.rgba image files
const SOURCE_PATH = 'C:\\Users\\vanca\\Desktop\\source\\sequence_short'

// Edit this to point to your ATEM
const ATEM_HOST = '192.168.1.5'

const fs = require('fs')
const path = require('path')
const { Atem } = require('atem-connection')

const allRGBAs = fs.readdirSync(SOURCE_PATH).filter(filename => {
	return filename.endsWith('.rgba')
}).map(filename => {
	return path.join(SOURCE_PATH, filename)
})

let frameNum = -1;
const provideFrame = function* () {
	const startTime = Date.now()
	for (const rgba of allRGBAs) {
		frameNum++
		console.log('Processing frame #%d (%s)', frameNum, rgba)
		yield fs.readFileSync(rgba)
	}

	const endTime = Date.now()
	console.log('Processing complete in %d seconds', (endTime - startTime) / 1000)
}

async function start() {
	const atem = new Atem();
	await atem.connect(ATEM_HOST)
	await sleep(1000) // Not sure why this is necessary, but it is...
	await atem.uploadClip(0, provideFrame(), 'clip0')
	process.exit(0)
}

async function sleep(duration) {
	return new Promise(resolve => {
		setTimeout(() => resolve(), duration)
	})
}

start().catch(error => console.error(error))
```

* **Other information**:
Support for async generators seems much more complex and would require a larger refactor of the data transfer system. It would also mean removing support for Node 8, as async generators only became available in Node 10.

Helps address https://github.com/nrkno/tv-automation-atem-connection/issues/110

### DIFF
--- a/src/atem.ts
+++ b/src/atem.ts
@@ -202,14 +202,7 @@ export class BasicAtem extends EventEmitter<AtemEvents> {
 
 			for (const commandName in DataTransferCommands) {
 				if (command.constructor.name === commandName) {
-					this.dataTransferManager.handleCommand(command).catch((e) => {
-						this.emit(
-							'error',
-							`Data transfer command handling failed: ${e}. Command: ${
-								command.constructor.name
-							} ${Util.commandStringify(command)}`
-						)
-					})
+					this.dataTransferManager.queueCommand(command)
 				}
 			}
 		}
--- a/src/atem.ts
+++ b/src/atem.ts
@@ -204,7 +204,14 @@ export class BasicAtem extends EventEmitter<AtemEvents> {
 
 			for (const commandName in DataTransferCommands) {
 				if (command.constructor.name === commandName) {
-					await this.dataTransferManager.handleCommand(command)
+					this.dataTransferManager.handleCommand(command).catch((e) => {
+						this.emit(
+							'error',
+							`MutateState failed: ${e}. Command: ${command.constructor.name} ${Util.commandStringify(
+								command
+							)}`
+						)
+					})
 				}
 			}
 		}
--- a/src/atem.ts
+++ b/src/atem.ts
@@ -172,7 +172,7 @@ export class BasicAtem extends EventEmitter<AtemEvents> {
 		const allChangedPaths: string[] = []
 
 		const state = this._state
-		commands.forEach((command) => {
+		commands.forEach(async (command) => {
 			if (state) {
 				try {
 					const changePaths = command.applyToState(state)
@@ -202,7 +202,7 @@ export class BasicAtem extends EventEmitter<AtemEvents> {
 
 			for (const commandName in DataTransferCommands) {
 				if (command.constructor.name === commandName) {
-					this.dataTransferManager.handleCommand(command)
+					await this.dataTransferManager.handleCommand(command)
 				}
 			}
 		})
@@ -671,15 +671,20 @@ export class Atem extends BasicAtem {
 		)
 	}
 
-	public uploadClip(index: number, frames: Array<Buffer>, name: string): Promise<DataTransfer> {
+	public uploadClip(
+		index: number,
+		frames: Iterable<Buffer> | AsyncIterable<Buffer>,
+		name: string
+	): Promise<DataTransfer> {
 		if (!this.state) return Promise.reject()
 		const resolution = Util.getVideoModeInfo(this.state.settings.videoMode)
 		if (!resolution) return Promise.reject()
-		const data: Array<Buffer> = []
-		for (const frame of frames) {
-			data.push(Util.convertRGBAToYUV422(resolution.width, resolution.height, frame))
+		const provideFrame = async function* (): AsyncGenerator<Buffer> {
+			for await (const frame of frames) {
+				yield Util.convertRGBAToYUV422(resolution.width, resolution.height, frame)
+			}
 		}
-		return this.dataTransferManager.uploadClip(index, data, name)
+		return this.dataTransferManager.uploadClip(index, provideFrame(), name)
 	}
 
 	public uploadAudio(index: number, data: Buffer, name: string): Promise<DataTransfer> {
--- a/src/atem.ts
+++ b/src/atem.ts
@@ -96,9 +96,7 @@ export class BasicAtem extends EventEmitter<AtemEvents> {
 
 		this.socket.on('commandsReceived', (commands) => {
 			this.emit('receivedCommands', commands)
-			this._mutateState(commands).catch((error) => {
-				this.emit('error', error)
-			})
+			this._mutateState(commands)
 		})
 		this.socket.on('commandsAck', (trackingIds) => this._resolveCommands(trackingIds))
 		this.socket.on('info', (msg) => this.emit('info', msg))
@@ -163,7 +161,7 @@ export class BasicAtem extends EventEmitter<AtemEvents> {
 		return this.sendCommands([command])[0]
 	}
 
-	private async _mutateState(commands: IDeserializedCommand[]): Promise<void> {
+	private _mutateState(commands: IDeserializedCommand[]): void {
 		// Is this the start of a new connection?
 		if (commands.find((cmd) => cmd.constructor.name === Commands.VersionCommand.name)) {
 			// On start of connection, create a new state object
--- a/src/atem.ts
+++ b/src/atem.ts
@@ -96,7 +96,9 @@ export class BasicAtem extends EventEmitter<AtemEvents> {
 
 		this.socket.on('commandsReceived', (commands) => {
 			this.emit('receivedCommands', commands)
-			this._mutateState(commands)
+			this._mutateState(commands).catch((error) => {
+				this.emit('error', error)
+			})
 		})
 		this.socket.on('commandsAck', (trackingIds) => this._resolveCommands(trackingIds))
 		this.socket.on('info', (msg) => this.emit('info', msg))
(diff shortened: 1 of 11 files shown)


baltedewit:
This is definitely a step in the right direction. I'm personally not a huge fan of generators but they seem to fit the requirements here. Not having asynchronous file loading feels like a big downside though. @Julusian I'm inclined to drop support for node 8 from here but remind me, what was the reason we could not last time?

Julusian:
master/v3.x has already dropped support for node 10 and lower, so might be a better target for this change. We kept it around for companion.
Ive never tried using generators myself, so dont have any opinions on them other than the syntax is weird (aka unfamiliar)
I think it needs to be async generators, sync concerns me as that can noticably impact performance for the rest of the application this is used in.

alvancamp:
Thanks for the feedback @baltedewit and @Julusian. I've gone ahead and made a first attempt at adding support for async generators. It seems to work for clips, but the unit test for WAV upload are failing now and I am unsure why. Any thoughts?

alvancamp:
After changing the base branch to `master` and fixing some linter errors, all tests now pass locally. I think this is ready for another review.

alvancamp:
Hm, sometimes when running my test file, I hit this `No frames available for transfer` error: https://github.com/alvancamp/tv-automation-atem-connection/commit/a3606a666f623ccac2bb498c052c201e8ffcb51d#diff-c7f99f09852f88bda4b59e740e320350eba365986c426e027a17af341263448eR30. There must be a race condition somewhere.

src/atem.ts
```diff
@@ -172,7 +172,7 @@ export class BasicAtem extends EventEmitter<AtemEvents> {
 		const allChangedPaths: string[] = []
 
 		const state = this._state
-		commands.forEach((command) => {
+		commands.forEach(async (command) => {
```
mint-dewit: I think this is better done as a for loop now because using async callbacks in a forEach may create race conditions.

baltedewit (approved):
Just a small comment regarding some async code, haven't tested it but looks good otherwise

src/atem.ts
```diff
@@ -202,10 +204,10 @@ export class BasicAtem extends EventEmitter<AtemEvents> {
 
 			for (const commandName in DataTransferCommands) {
 				if (command.constructor.name === commandName) {
-					this.dataTransferManager.handleCommand(command)
+					await this.dataTransferManager.handleCommand(command)
```
Julusian: Balte is correct about the for loop, but I dont think we should be awaiting this here.

This will allow us to be running multiple _mutateState calls in parallel if this await takes any time, which means we could handle commands out of order, and make a mess of our state object.

I think it would be better to make the DataTransferManager have a p-queue for handling commands to limit concurrency of the handlers inside there without risking interfering with the main state object updating

src/dataTransfer/dataTransfer.ts
```diff
@@ -35,8 +35,10 @@ export default abstract class DataTransfer {
 		return this.completionPromise
 	}
 
-	public abstract start(): Commands.ISerializableCommand[]
+	public abstract start(): Commands.ISerializableCommand[] | Promise<Commands.ISerializableCommand[]>
```
Julusian: I would make all these methods always return the promise version. The implementations can be made async to handle the change really easily, and it keeps it simpler

src/dataTransfer/dataTransferClip.ts
```diff
@@ -50,11 +70,25 @@ export default class DataTransferClip extends DataTransfer {
...
 
-	public gotLock(): Commands.ISerializableCommand[] {
+	public gotLock(): Promise<Commands.ISerializableCommand[]> {
 		this.state = Enums.TransferState.Locked
 		return this.start()
 	}
+
+	private waitForStart(): Promise<void> {
+		return new Promise((resolve) => {
+			setInterval(() => {
```
Julusian: Instead of an interval I would use a promise as a semaphore.
I can dig out an example for you if wanted

src/dataTransfer/index.ts
```diff
@@ -105,10 +105,11 @@ export class DataTransferManager {
 			}
 		}
 		if (command.constructor.name === Commands.DataTransferErrorCommand.name) {
-			lock.transferErrored(command.properties.errorCode)
+			void lock.transferErrored(command.properties.errorCode)
```
Julusian: Is void appropriate? Could it result in an unhandledPromiseRejection ?

src/dataTransfer/dataTransferClip.ts
```diff
@@ -5,41 +5,61 @@ import DataTransferFrame from './dataTransferFrame'
...
+		this.curFrame.state = Enums.TransferState.Locked
+		commands.push(...this.curFrame.start())
+		this.started = true
 		return commands
 	}
 
-	public handleCommand(command: Commands.IDeserializedCommand): Commands.ISerializableCommand[] {
+	public async handleCommand(command: Commands.IDeserializedCommand): Promise<Commands.ISerializableCommand[]> {
+		if (!this.started) {
+			await this.waitForStart()
```
alvancamp, line 42: Is this creating a race condition? Does this guarantee that commands will be handled in the order they were received? If not, how can that guarantee be made?

mint-dewit: Iirc the protocol uses an internal counter to ensure packets are received in order (by packetId), a race condition however may still occur because the commands are emitted as events so while 1 command is being handled, another may bubble up and due to async code create a race condition. It seems a little far fetched but to be certain we could use a promise queue to ensure handling commands in order?

Julusian: It might be, but im not sure its necessary to wait for the start here.
Until the commands returned from start() are sent then anything fed into handleCommand cant be for this transfer, and this.started will be made true before any of those commands are sent. So I think this should either be a `return []` or maybe error in some way but im not sure if that should error back to the atem or log something

src/dataTransfer/dataLock.ts
```diff
@@ -36,17 +36,18 @@ export default class DataLock {
 
 			if (this.isLocked) {
 				// TODO - this flow should never be hit
-				this.lockObtained()
+				void this.lockObtained()
```
Julusian: Is this void appropriate? what happens if it throws?

Julusian (changes requested):

src/dataTransfer/dataTransferClip.ts
alvancamp: Yes please, I'd appreciate an example.

src/dataTransfer/dataTransferClip.ts
alvancamp: In my previous testing I found that it was necessary to wait for start, because some lock-related commands were coming in to `handleCommand` before `start` had finished executing. However, since adding a `p-queue` to the `DataTransferManager`, I've been unable to replicate that issue and have removed this code for now.

src/atem.ts
alvancamp: Okay, I added a 'p-queue` to `DataTransferManager`. Hopefully I interpreted your request correctly. Let me know if I missed the mark.

src/dataTransfer/dataTransfer.ts
alvancamp: Done.

src/atem.ts
Julusian: that looks like what I was hoping for :)

src/atem.ts
```diff
@@ -161,7 +163,7 @@ export class BasicAtem extends EventEmitter<AtemEvents> {
 		return this.sendCommands([command])[0]
 	}
 
-	private _mutateState(commands: IDeserializedCommand[]): void {
+	private async _mutateState(commands: IDeserializedCommand[]): Promise<void> {
```
Julusian: This no longer needs to be async, so would be good to remove that

src/dataTransfer/index.ts
```diff
@@ -61,68 +64,95 @@ export class DataTransferManager {
...
-		if (
-			command.constructor.name === Commands.LockObtainedCommand.name ||
-			command.constructor.name === Commands.LockStateUpdateCommand.name
-		) {
-			lock = allLocks[command.properties.index]
-		} else if (typeof command.properties.storeId === 'number') {
-			lock = allLocks[command.properties.storeId]
-		} else if (command.properties.transferId !== undefined || command.properties.transferIndex !== undefined) {
-			for (const _lock of allLocks) {
+	public async handleCommand(command: Commands.IDeserializedCommand): Promise<void> {
```
Julusian: This method doesnt need to be async anymore either.
I think it would be good to rename to queueCommand, to make it clear to callers that it queues things instead of processing immediately

Julusian (commented):
Just a couple of tiny style thing :)
Other than that I am happy

PR #123, opened 2021-12-07
feat: generate multiviewer labels

### LINKED ISSUE #59: Updating input name doesn't update multiviewer
**I'm submitting a ...**
[*] bug report

**Affected**
Constellation in v8.0.2
TV Studio HD in v7.5.2
2ME in v8.0.2

**Unaffected**
Atem Mini Pro ISO v8.6.4
Atem Mini Extreme ISO v8.6.4
(It is assumed that all of the mini range are unaffected by this)

**Summary**
Using `setInputSettings` to set the longName of a source does not update the multiviewer.
Other connected clients picks up the new device name, so the set command does work successfully. But atem client is sending an encoded image with the new name, which we are not doing.

**Notes**
It is the same process as a stills upload, with the following differences:
storeId = 0xffff
mode (TransferMode) = 0x0201
transferIndex = sourceId
filename = 'Label'
description = ''
hash = 0x00

The image appears to be 8bit per pixel, with some unknown encoding. Viewing in greyscale looks weird, but is recognisable.
RLE encoding can be used during the upload.
For one model, the image is 640 x 190, and contains a render for each of 4k?, 1080, and 720. But this varies based on device capabilities.

**WARNING**
We found that sending images with a full gradient of grey would crash the ATEM, so this needs extensive testing to ensure that our implementation is safe.

**Other information**
Wireshark of example [atem-set-source-name-does-upload.zip](https://github.com/nrkno/tv-automation-atem-connection/files/3653143/atem-set-source-name-does-upload.zip)
Image extracted from wireshark capture: [ripped-from-wireshark.zip](https://github.com/nrkno/tv-automation-atem-connection/files/3653190/ripped-from-wireshark.zip)

Julusian:

feature
closes #59

When setting the longName for an input, the multiviewer is not updated

There is a new method to generate and upload the label image for the multiviewer. This uses freetype2 to render the text, with a similar drawing style to the default atem rendering. 

This has been left as a separate method for now to avoid complex logic to figure out when or if the images should be uploaded.
There is also a small risk that the upload could crash an atem if it contains a bad bitmap, so it is safer to make it opt-in.

There is also a method to upload a custom buffer, more documentation is needed on how that buffer should look

* **Other information**:


Julusian:
PXL_20220328_222624854
Top is this library, bottom is original

I'm pretty happy now. Not 100% perfect, but its getting close enough. I don't think we can use the original font due to licensing (I think it might be Helvetica), but this free font is close

Julusian:
Im not 100% happy with the error handling for dataTransfers, but I don't think it is any worse than before

PR #130, opened 2022-10-22
feat: streaming audio bitrates

bevand10:
### What kind of change does this PR introduce?
Support for the `STAB` message -
`ST`reaming
`A`udio
`B`itrates

### What is the current behaviour?
`STAB` is not currently supported.

#### `STAB` message receipt is supported
Data received is added to the `atem.streaming` object. Here's the results of a `console.log(atem.state);` command in Node when using this branch.
```
...
  streaming: {
    service: {
      serviceName: 'my service name',
      url: 'rtmp://a.rtmp.youtube.com/live2',
      key: '****-****-****-****-****',
      bitrates: [Array]
    },
    audioBitrates: { lowBitrate: 128000, highBitrate: 128000 },
    status: { state: 1, error: 0 },
    duration: { hours: 0, minutes: 0, seconds: 0, frames: 0, isDropFrame: false },
    stats: { encodingBitrate: 0, cacheUsed: 0 }
  },
...
```
#### `STAB` message transmission is supported
A new command has been added to set streaming audio bitrates. BMD documentation states that a STAB message defines two bitrate values, one for framerates up to 30fps, and another for framerates greater than 30fps. It refers to these as `lowBitrate` and `highBitrate` values. Strictly speaking, IMO, these should really be referred to as `lowFramerates` and `highFramerates` or even more explicitly, `upto30fps` and `morethan30fps`, but I'll overlook that mistake from BMD!

This example sets both values to 128k:
```
atem.setStreamingAudioBitrates(128000, 128000);
```

#### Safe defaults
When calling `atem.setStreamingAudioBitrates()` (i.e. without params), safe defaults are applied:
```
    audiobitrates: { lowBitrate: 128000, highBitrate: 192000 },
```

image


Julusian:
Thanks, this is published in v3.1.0

PR #131, opened 2022-11-21
fix: timing issue and socket leak

ianshade:
Bug fix in the `AtemSocketChild`

1. due to using `Date.now()`, some calculations are prone to issues arising from system clock jumps (NTP sync etc.)
2. dgram sockets are leaking when reconnecting: a socket is created, but the previous one is not closed - **needs confirmation whether I'm not mistaken**

The bugs are fixed by:
1. uses `performance.now()` to rely on the monotonic clock instead of the system clock
2. closes the old socket when creating a new one; stops some timers during the period when `onDisconnect` is awaited; stops the timers when `disconnect()` is called; does not create new socket when `disconnect()` is called

* **Other information**:


src/lib/atemSocketChild.ts
```diff
@@ -191,12 +195,29 @@ export class AtemSocketChild {
...
 			trackingId,
-			lastSent: Date.now(),
+			lastSent: performance.now(),
 			payload: buffer,
 			resent: 0,
 		})
 	}
 
+	private _recreateSocket(): Socket {
+		this._closeSocket().catch((err) => {
```
Julusian: This won't ever catch anything, due to how `_closeSocket` is implemented

src/lib/atemSocketChild.ts
```diff
@@ -349,7 +370,7 @@ export class AtemSocketChild {
 			for (let i = fromIndex; i < this._inFlight.length; i++) {
 				const sentPacket = this._inFlight[i]
 				if (sentPacket.packetId === fromId || !this._isPacketCoveredByAck(fromId, sentPacket.packetId)) {
-					sentPacket.lastSent = Date.now()
+					sentPacket.lastSent = performance.now()
```
Julusian: I remember someone said that `performance.now()` is slow. Perhaps we should do `const now = performance.now()` outside of the loop, to avoid calling it inside these short loops. Same goes for `_checkForRetransmit`.
It might also want a guard `if (this._inFlight.length) return`, to avoid calling it unnecessarily

Julusian (approved):

PR #134, opened 2022-12-05
ci: Update depreciated actions

PeterC89:
Updates GHA workflows to remove depreciated calls

Workflows run successfully

Workflows continue to run successfully with the same behaviour

* **Other information**:

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -69,7 +69,7 @@ jobs:
       - name: Bump version and build
         if: ${{ steps.do-publish.outputs.publish }}
         run: |
-          PRERELEASE_TAG=nightly-${{ github.ref_name }}
+          PRERELEASE_TAG=nightly-$(echo "${{ github.ref_name }}" | sed -r 's/[^a-z0-9]+/-/gi')
           yarn release --prerelease $PRERELEASE_TAG
           yarn build
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yaml
```diff
@@ -55,26 +55,21 @@ jobs:
...
-      - name: Get the Prerelease tag
-        id: prerelease-tag
-        uses: yuya-takeyama/docker-tag-from-github-ref-action@2b0614b1338c8f19dd9d3ea433ca9bc0cc7057ba
-        with:
-          remove-version-tag-prefix: false
       - name: Bump version and build
         if: ${{ steps.do-publish.outputs.publish }}
         run: |
-          PRERELEASE_TAG=nightly-${{ steps.prerelease-tag.outputs.tag }}
+          PRERELEASE_TAG=nightly-${{ github.ref_name }}
```
Julusian:
```suggestion
          PRERELEASE_TAG=nightly-$(echo "${{ github.ref_name }}" | sed -r 's/[^a-z0-9]+/-/gi')
```
I think this will need to be like what we have in tsr. Either tsr was doing overkill processing, or we don't use branches enough in this lib to hit the issue with attempting to publish with a `/` in the version

Julusian (approved):

mint-dewit (approved):

jstarpl (approved):

PR #147, opened 2023-07-21
fix: prevent 'Lost lock mid-transfer' error with multiple transfers

ianshade:
bug fix 

Uploading multiple stills at once or one after the other sometimes leads to the `Lost lock mid-transfer` error
https://github.com/nrkno/sofie-atem-connection/issues/143

The idea is to make the locking more robust by:
  - unlocking only when the queue is empty,
  - aborting transfers when an unexpected unlock occurs, but not when a transfer gets queued between sending the unlocking LockStateCommand and receiving LockStateUpdateCommand back

* **Other information**:

### DIFF
--- a/src/dataTransfer/dataTransferQueue.ts
+++ b/src/dataTransfer/dataTransferQueue.ts
@@ -270,17 +270,20 @@ export class DataTransferLockingQueue extends DataTransferQueueBase {
 
 	/** The status of the lock has changed */
 	public updateLock(locked: boolean): void {
+		const wasLocked = this.isLocked
 		this.isLocked = locked
 
-		if (!locked) {
+		if (wasLocked && !locked) {
+			// only abort when it becomes unlocked unexpectedly, otherwise we're likely waiting to
+			// obtain lock for a transfer queued shortly after sending unlock command
 			this.tryAbortTransfer(new Error('Lost lock mid-transfer'))
 		}
 	}
 
 	protected transferCompleted(): void {
-		if (this.isLocked) {
-			// Make sure that we don't try to start the next before the unlock completes
-			// TODO - is this durable?
+		if (this.isLocked && !this.taskQueue.length) {
+			// unlock only when queue is empty
+
 			this.isLocked = false
 
 			debug(`Completing transfer`)


Julusian (commented):
I think this looks good, but I want to give it a bit of a stress test before merging it.

I have a feeling like I had it doing this behaviour of not unlocking between uploads once, but that something else was expecting it to be unlocked between each, causing one of them to get stuck or fight over the lock.

Julusian:
Yeah this is definitely behaving a lot better than before. It appears that previously it would sometimes get stuck for some reason..

There is still an easy to reproduce case of uploads getting stuck, but I expect that is not really related to this change.

Using:
```
const { Atem } = require('../dist')
const fs = require('fs')

const file = fs.readFileSync('./testframe.rgba')

if (process.argv.length < 3) {
	console.error('Expected ip address as parameter')
	process.exit(1)
}
const IP = process.argv[2]

const conn = new Atem({})

function uploadBatch() {
	const count = 8
	Promise.resolve().then(async () => {
		console.log('starting batch')

		const ps2 = []
		for (let i = 0; i < count; i++) {
			ps2.push(conn.clearMediaPoolStill(i))
		}
		await Promise.allSettled(ps2)
		console.log('batch cleaned')

		const ps = []
		for (let i = 0; i < count; i++) {
			ps.push(uploadNext(i))
		}

		await Promise.allSettled(ps)
		console.log('batch done')

		setTimeout(() => uploadBatch(), 500)
	})
}

async function uploadNext(myNumber) {
	const stuckTimeout = setTimeout(() => {
		console.log('')
		console.log('UPLOAD GOT STUCK')
		console.log('')
	}, 20000)

	const uploadStarted = Date.now()
	// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
	await conn.uploadStill(myNumber % conn.state.media.stillPool.length, file, 'contemplation..', '').then(
		(_res) => {
			console.log(`Uploaded still #${myNumber} in ${Date.now() - uploadStarted}ms at 1080p`)

			clearTimeout(stuckTimeout)
		},
		(e) => {
			console.log(`fail from #${myNumber}`, e)

			clearTimeout(stuckTimeout)
		}
	)
}

conn.on('error', console.log)
conn.on('disconnected', () => {
	console.log('Connection lost')
	process.exit(0)
})
conn.connect(IP).catch((e) => {
	console.error(e)
	process.exit(0)
})
conn.once('connected', () => {
	console.log(`connected`)
	uploadBatch()
})
```
You might need to increase the `count` to get it to hit the scenario.
For this it is important to also have the atem software open (for me on another machine, should be fine on the same)
It will upload the first batch fine, with the atem software waiting for the lock to be released (orange circle shows over the thumbnails)  
The batch completes uploading, the atem software starts downloading, then the upload of the second batch begins, and fails due to not having the lock (`Error: Unknown error 5`) (if the download completes first this does not happen, tune the count to get them to overlap)

Trying the same script before this change, it fails to even upload the first batch, with the second image failing due to the lock lost error.

@ianshade I am happy to merge this as is, unless you want to look into this likely unrelated other locking issue first?

Julusian (approved):

Issue #99, opened 2021-01-01
Read audio signal levels


SNRSE:
* **I'm submitting a ...**
[ ] bug report
[ ] feature request
[ ] question about the decisions made in the repository
[ X] question about how to use this project

* **Summary**
I want to listen to the Atem so i can get the current level of an audio input.

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. StackOverflow, personal fork, etc.)

Julusian:
Are you wanting the current signal level, or the current fader level?
For fairlight or classic audio?

I dont think this library has any support for signal levels yet, but fader levels are all on the state object

SNRSE:
Hi yes i want the current Signal Level to do some autoswitching.

Julusian:
yeah, so this isnt currently supported. No reason other than it hasnt been requested/needed before.

It won't be too hard to do, just a bit of thought to now throw out too many updates that could impact other code which listens on the state

aaronpk:
i’m also super interested in this feature! Since it’s likely a relatively uncommon use case I wouldn’t mind having to explicitly opt in to get the audio level events from the library to avoid flooding everyone else.

imorrish:
For Video Follow Audio (which is why this is being requested), you could get away with having "Peak" value in the state and a Clear Peak method. This would make it easy for apps to detect when a threshold audio level had been reached and clear the peak value once they had switched input based on peak level.

keith8496:
I would love to see audio signal levels implemented. For my use-case, I'm trying to build a web-based dashboard to bring numerous pieces of our livestream workflow together in once place.

**It looks like audio levels have been added to LibAtem since this issue was last discussed.**

I believe this turns the audio level feature on and off. It's opt-in. Off by default.
https://github.com/LibAtem/LibAtem/blob/master/LibAtem/Commands/Audio/Fairlight/FairlightMixerSendLevelsCommand.cs

This looks like it receives the master levels:
https://github.com/LibAtem/LibAtem/blob/master/LibAtem/Commands/Audio/Fairlight/FairlightMixerMasterLevelsCommand.cs

And the source levels:
https://github.com/LibAtem/LibAtem/blob/master/LibAtem/Commands/Audio/Fairlight/FairlightMixerSourceLevelsCommand.cs

[...]

PR #148, opened 2023-07-22
Feat/fairlight mixer send levels

keith8496:
Feature - Adds Fairlight audio levels

[ Read audio signal levels #99 ](https://github.com/nrkno/sofie-atem-connection/issues/99)

Adds two new commands: "startFairlightMixerSendLevels" and "stopFairlightMixerSendLevels".
Adds a new event type: "levelChanged".

* **Other information**:

Audio levels are opt-in via start and stop commands. This is consistent with LibATEM and ATEM Software Control. Default to off.

I haven't looked at audio level support for classic audio. I don't have an ATEM device to test against. I am testing Fairlight audio against an ATEM 1 M/E Constellation HD.

The lead devs should discuss whether they want audio levels sent in the existing "stateChanged" event or a new "levelChanged" event. I've tested both methods. The event fires many times per second, so I created the event "levelChanged" to avoid sending the whole state over-and-over again. I'm happy to use "stateChanged" if that's what the devs prefer.

From issue #99:

> I would love to see audio signal levels implemented. For my use-case, I'm trying to build a web-based dashboard to bring numerous pieces of our livestream workflow together in once place.
> 
> It looks like audio levels have been added to LibAtem since this issue was last discussed.
> 
> I believe this turns the audio level feature on and off. It's opt-in. Off by default.
> https://github.com/LibAtem/LibAtem/blob/master/LibAtem/Commands/Audio/Fairlight/FairlightMixerSendLevelsCommand.cs
> 
> This looks like it receives the master levels:
> https://github.com/LibAtem/LibAtem/blob/master/LibAtem/Commands/Audio/Fairlight/FairlightMixerMasterLevelsCommand.cs
> 
> And the source levels:
> https://github.com/LibAtem/LibAtem/blob/master/LibAtem/Commands/Audio/Fairlight/FairlightMixerSourceLevelsCommand.cs
> 
> FWIW, this has been implemented in PyATEM, although I think he's got a bug in his conversion from raw data to decibels. I've been experimenting with his project. The additional load added by the audio levels seems insignificant on a 2013 MacMini running Ubuntu Server 22.04 LTS.
> https://github.com/git-2-github/martijnbraam-pyatem

> I have an experimental branch with Fairlight audio signals for the devs' consideration. It's "alpha" quality at this point, not pull request ready. This is my first time working with Typescript and my first time building with Yarn.
> 
> I used the LibAtem resources described above. Audio levels are opt-in with commands to turn them on and off. This is consistent with LibAtem and I believe the official SDK.
> 
> I've tried two versions of this. The first version adds audio levels to state under fairlight.master.levels and fairlight.inputs.[i].sources.[s].levels. This triggers stateChanged events many times a second (as you would expect). It might not be a good option. The second version moves audio levels to their own levelChanged event. At this point I figure I should stop and ask if the devs if they have any feedback.
> 
> https://github.com/keith8496/sofie-atem-connection/tree/feat/FairlightMixerSendLevels

> Hi @keith8496! Thanks for looking into this!
> Could you open a pull request from that branch? That way it's easier for us to provide feedback on the code.
> 
> We might be a bit slow replying during the next few weeks as most of the devs are on vacation.


src/commands/Fairlight/FairlightMixerMasterLevelsCommand.ts
```diff
@@ -0,0 +1,52 @@
...
+		}
+
+		if (!state.fairlight.master) {
+			throw new InvalidIdError('Fairlight.Master')
+		}
+
+		state.fairlight.master.levels = this.properties
+
+		return `fairlight.master.levels`
+	}*/
```
Julusian:
```suggestion
```

src/commands/Fairlight/FairlightMixerMasterLevelsCommand.ts
```diff
@@ -0,0 +1,52 @@
+import { FairlightAudioMasterLevelsState } from '../../state/fairlight'
+//import { AtemState, InvalidIdError } from '../../state'
```
Julusian:
```suggestion
```

src/commands/Fairlight/FairlightMixerSourceLevelsCommand.ts
```diff
@@ -0,0 +1,70 @@
+import { /*FairlightAudioInput,*/ FairlightAudioSourceLevelsState } from '../../state/fairlight'
+//import { AtemState, InvalidIdError } from '../../state'
```
Julusian:
```suggestion
import { FairlightAudioSourceLevelsState } from '../../state/fairlight'
```

src/commands/Fairlight/FairlightMixerSourceLevelsCommand.ts
```diff
@@ -0,0 +1,70 @@
...
+		}
+
+		const sourceIdStr = this.source.toString()
+		const source = input.sources[sourceIdStr] || {}
+		input.sources[sourceIdStr] = source
+
+		source.levels = this.properties
+
+		return `fairlight.inputs.${this.index}.sources.${sourceIdStr}.levels`
+	}*/
```
Julusian:
```suggestion
```

src/atem.ts
```diff
@@ -65,6 +67,7 @@ export type AtemEvents = {
 	connected: []
 	disconnected: []
 	stateChanged: [AtemState, string[]]
+	levelChanged: [FairlightAudioMasterLevelsState | FairlightAudioSourceLevelsState, string]
```
Julusian: I think the parameters of this need changing. Currently it wont be possible to differentiate different sources with the same index, and this doesnt play particularly well with typescript. Once you have checked the string is or isnt 'master', you will have to cast the object to the correct type.
 
I think the best solution would be to make it be a single parameter, which wraps the object. That will additionally make it nice and easy to narrow the type of the object in typescript.

Something like:
```
interface FairlightAudioMasterLevels {
  system: 'fairlight'
  type: 'master'
  levels: FairlightAudioMasterLevelsState
}

interface FairlightAudioSourceLevels {
  system: 'fairlight'
  type: 'source'
  index: number
  source: bigint
  levels: FairlightAudioSourceLevelsState
}
```

Julusian (changes requested):
Overall it looks good

I think a separate levelsChanged event is a good idea. That will help avoid this becoming an accidental performance issue in applications. (as evidenced by the bug in the Constellation HD which emits some state changes on every frame, which causes issues for some users of this library)

PR #149, opened 2023-07-24
perf: Add RLE encoding

ianshade:
Performance improvement for uploading stills and clips that benefit from RLE

Frames are sent uncompressed, which consistently takes about 600ms per 1920x1080 frame

Encodes still and clip frames with Run-Length Encoding supported by ATEM switchers, making some frames (those that have lots of solid color areas) upload significantly faster (even down to 100ms).
Also increases `MAX_PACKETS_TO_SEND_PER_TICK` to 50 which seems to be the sweet spot.

* **Other information**:
This currently also includes changes from https://github.com/nrkno/sofie-atem-connection/pull/147

  On my 12th Gen Intel notebook CPU encoding takes about 30-40ms for a 1920x1080 frame, no matter how compressible it is. Should we try WASM with SIMD for the color space conversion and RLE to squeeze out some more performance? Should we make RLE optional? - I'm open for suggestions.
  
  The code is ready for review, but some help is needed to fix existing tests that I believe have wrong hashes in them.


Julusian:
Nice, this is one of things that I have been wondering about, but the benefit was too much of an unknown for me to be able to motivate.

I'll give this and your other PR a proper read next week.

> Should we make RLE optional?

I think it would be good to make it opt out. Perhaps always disabled when running in single-threaded mode, as that 30-40ms of cpu time might be long enough for the connection to drop?  
Making it opt-out gives a nice escape hatch in case it does cause connection issues, or for when you know that the RLE will do nothing (eg, source is from a jpeg).  
I think this should be specified with an options object, so that we can add some other options later, like specifying the input format or output colourspace. (we currently assume its a RGBA buffer, and converting to 709)

> Should we try WASM with SIMD for the color space conversion and RLE to squeeze out some more performance?

If we were to move the RGB->YUV into that too then WASM will probably be beneficial. The problem with WASM is that the pixels will need to be copied from a nodejs buffer into a WASM buffer, and the inverse on the way out, so if not careful it could be more costly.  
I prototyped a similar conversion in my streamdeck library, which did halve the time of the buffer operations, but the numbers (for 72x72 pixels) were small enough to make the benefit questionable

src/dataTransfer/dataTransferUploadBuffer.ts
```diff
@@ -94,19 +99,29 @@ export abstract class DataTransferUploadBuffer extends DataTransfer<void> {
...
 		// Take a little less because the atem does that?
-		const chunkSize = props.chunkSize - 4
+		// const chunkSize = props.chunkSize - 4
+		const chunkSize = Math.floor(props.chunkSize / 8) * 8
 
 		for (let i = 0; i < props.chunkCount; i++) {
 			// Make sure we don't end up with an empty slice
 			if (this.#bytesSent >= this.data.length) break
 
+			let shortenBy = 0
```
Julusian, line 124: It would be nice to have some comments here, explaining what this block is intending to do

Julusian (commented):
If you could make it opt-out as described in https://github.com/nrkno/sofie-atem-connection/pull/149#issuecomment-1648338229, then I am happy.

I am open to doing some wasm, that might be better as a follow up PR? I don't mind. (I am tempted to implement that myself)

Julusian:
I have made this opt-out, and this is now released.

I also started playing around with doing the RGBA->YUVA conversion in a native library in #152, we could look at moving the RLE to that too later on, once/if that gets merged

PR #154, opened 2023-09-25
feat: remove use of nanotimer

Julusian:

feature

`nanotimer` is used to ensure accurate timing in timers, at the cost of increased cpu usage.

@mint-dewit can you remember any justification for using nanotimer over regular nodejs timers? https://github.com/nrkno/sofie-atem-connection/commit/84482cad64bf5213eef1ce99eaaecc2e36762ae8 I can't find anything in asana or the github issues that give an explanation

use `setTimeout` instead of `nanotimer`.

In my testing, this gave accurate enough results. And as we run this timer in a worker thread, there shouldnt be much/any contention on cpu time within the thread. It is possible that the timer callback will be late if overall system cpu usage is high, but I am not sure if nanotimer would do this much better, as it will be compounding the cpu starvation with its spin loop, and it still relies on the event loop running to run the next iteration.

* **Other information**:
Testing shows `setTimeout` is able to pretty accurately fire after short periods. This will improve performance, as nanotimer spins the cpu to hit its target. Tested using a variation of https://github.com/nodejs/node/issues/26578#issuecomment-471392004

This could probably do with some more stress testing in a production like environment to ensure it doesnt cause any issues in a real environment.

## Summary

This PR removes the nanotimer dependency and replaces its use with Node.js built-in timers (setTimeout / setInterval). Timers are managed in a worker thread to reduce contention. The change reduces CPU spin-loop overhead from nanotimer while preserving acceptable short-period timing for the ATEM protocol; the author recommends further production-like stress testing because very high system CPU could still cause late callbacks.

## Changes

**package.json**
-	Removed `nanotimer` dependency

**src/@types/nanotimer.d.ts**
-	Removed TypeScript declaration file for the `nanotimer` module

**src/lib/atemSocketChild.ts**
-	Replaced NanoTimer usages with standard Node.js timers and updated field types:
-		`_reconnectTimer` → NodeJS timer (setInterval)
-		`_retransmitTimer` → NodeJS.Timeout (setTimeout)
-		`_ackTimer` → NodeJS.Timeout (setTimeout)
-	Added `_triggerRetransmitTimer()` helper to create/clear a retransmit setTimeout only when packets are inflight
-	Refactored ACK logic:
-		Use a 5ms setTimeout to coalesce ACKs; send immediate ACK when MAX_PACKET_PER_ACK is reached
-		Clear `_ackTimer` where appropriate (socket close, ack send)
-	Call `_triggerRetransmitTimer()` on packet send and on ack receipt to keep retransmit scheduling aligned with inflight state
-	Updated all timer cleanup to use `clearTimeout` / `clearInterval`

## Testing

-	Short-period timing tests (based on a Node.js issue comment) showed setTimeout provides sufficient accuracy and reduced CPU usage versus nanotimer.
-	Author recommends stress testing in production-like environments to validate behavior under high CPU load.

## Notes / Review comments

-	codecov reported patch coverage 95.45% with 1 changed line missing coverage (file: src/lib/atemSocketChild.ts).
-	Reviewer noted historical concerns that Node timers might have missed tight ACK timing windows (~5ms) in older Node versions; current tests suggest tolerances and/or Node timer accuracy have improved, but this should be validated under realistic load.


mint-dewit:
> can you remember any justification for using nanotimer over regular nodejs timers?

My memory is probably incorrect given how long ago it was but I think the default timers were not always fast enough to send the acknowledgements. I think at that time the atem protocol was quite finicky and it was required to send the ack's within 5 ms or something. Being too slow would cause the entire connection to be thrown away by the atem. I think they've made the protocol quite a bit more tolerant to timing these days.

It's also possible node has just gotten more accurate with it's timers, after all 2018 is a long time ago.

mint-dewit (approved):

PR #157, opened 2023-12-01
feat: command batching SOFIE-2549

Julusian:

feature

When sending commands to the atem, it is only possible to send one per packet.

The `sendCommands` method is now exposed, allowing for sending multiple commands in one packet. 
Each call to the method will result in packets being sent, no buffering/debouncing is performed between calls. Commands are packed into packets in order, up to a defined byte limit per packet.

The maximum packet size is defined based on the `packetMtu` property supplied to the constructor. Care should be taken when setting this, as any upload transfers do not fully respect this value and will result in packets failing to send.

* **Other information**:

This partially covers https://github.com/nrkno/sofie-atem-connection/issues/140, but only solves the case of constructing the commands manually (or with `atem-state`), not usage through the methods provided on the `Atem` class.  
A friendlier batching api can be built on top of this work by someone later.

I have tested this against a Constellation 2ME HD by sending a pair of `MixEffectKeyDVECommand` commands multiple time at once, enough so that it spans multiple packets. The packets were acked, indicating that they were transmitted and received successfully.

### DIFF
--- a/src/lib/packetBuilder.ts
+++ b/src/lib/packetBuilder.ts
@@ -31,11 +31,12 @@ export class PacketBuilder {
 
 		const totalLength = payload.length + 8
 		if (totalLength > this.#maxPacketSize) {
-			throw new Error(`Comamnd ${cmd.constructor.name} is too large for a single packet`)
+			// Command is too large for a normal packet, try sending it on its own anyway
+			this.#finishBuffer(totalLength)
 		}
 
 		// Ensure the packet will fit into the current buffer
-		if (totalLength + this.#currentPacketFilled > this.#maxPacketSize) {
+		if (totalLength + this.#currentPacketFilled > this.#currentPacketBuffer.length) {
 			this.#finishBuffer()
 		}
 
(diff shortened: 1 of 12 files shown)


src/lib/packetBuilder.ts
```diff
@@ -0,0 +1,68 @@
...
+		if (typeof cmd.serialize !== 'function') {
+			throw new Error(`Command ${cmd.constructor.name} is not serializable`)
+		}
+
+		const rawName: string = (cmd.constructor as any).rawName
+		const payload = cmd.serialize(this.#protocolVersion)
+
+		const totalLength = payload.length + 8
+		if (totalLength > this.#maxPacketSize) {
+			throw new Error(`Comamnd ${cmd.constructor.name} is too large for a single packet`)
```
Julusian: I'm not sure if this is a good idea. Perhaps instead it should be put as its own packet, so that it stands a chance of being accepted by the atem?

nytamin: Agreed, the default behaviour should probably be to put it into a new packet.

Would it be worth adding another argument into `sendCommands()` to allow for a consumer to get notified if the commands doesn't fit into one packet?
```typescript
export class AtemSocket {
// ...
/** 
 * sendCommands
 * @param commands commands to send
 * @param onlyOnePacket ensure commands are sent as a single packet. (throws if they doesn't fit)
 */
sendCommands(commands: Array<ISerializableCommand>, onlyOnePacket?: boolean)
```

nytamin (approved):
Looks good to me from an initial read through

src/lib/packetBuilder.ts
Julusian: I'm not sure if a new parameter will be useful. These large commands will always be coming from the 'DataTransfer' system, so either way its behaviour will end up pretty much hardcoded one way.

But I think its a question of whether the transfer should immediately fail due to the packets being unsendable (I don't know how/if the logic will handle this anyway), or if they should be sent and left to timeout

Julusian: I think I prefer the idea of sending them anyway. I dont expect this to be a problem in practise (I dont know why anyone would need to change the maxPacketSize), so if it does then it can be dealt with if it does become a problem.

PR #171, opened 2025-03-08
fix: use the official freetype2 package

aminya:
This updates the freetype dependency to freetype2. All the changes from the fork have been merged and new fixes have been added.

## About the Contributor

Personal contribution

## Type of Contribution

Dependency update

## Testing Instructions

The current tests should pass

## Status

- [x] Relevant documentation (code comments, [system documentation](https://nrkno.github.io/sofie-core/)) has been added / updated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "debug": "^4.3.4",
     "eventemitter3": "^4.0.7",
     "exit-hook": "^2.2.1",
-    "freetype2": "^2.0.0",
+    "freetype2": "^2.1.0",
     "nanotimer": "^0.3.15",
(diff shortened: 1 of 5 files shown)


nytamin:
I've checked a bit with the team, and I'm leaning towards keeping the existing dependency.

It seems like there are a few issues in the official package related to it's usage of `node-gyp` and `node-addon-api`, as those dependencies are too new for node 14 (as reported in their engines), so merging this probably should be a semver major change (or reported as a bug, the package does claim to support 14).

Is there any particular issue you'd like to resolve with changing the dependency to the official one?

PR #178, opened 2025-09-08
feat: Run order groups for ATEM commands

rjmunro:
## About the Contributor

This pull request is posted on behalf of the BBC.

## Type of Contribution

This is a Feature

It's required for the following bugfix: https://github.com/Sofie-Automation/sofie-atem-state/pull/37.

## Current Behavior

Commands sent to an ATEM are batched where possible. You cannot control which batch they go in. Additionally the ATEM firmware sometimes seems to execute things out of order within a batch.. On some occasions, order is important, for example when setting up and switching on a key it's important not to switch it on until after it's input are all set. Also when switching it off, it's important not to make any other changes until after it has switched off. 

## New Behavior

You can set a runOrderGroup property on an ATEM command. This is a number. Commands will only be batched if they are in the same group. Groups will be run in numerical order. Default is 0, so you can use a positive number to run something after other commands (such as turning on a key after sending it's settings), or a negative number to ensure it happens first (such as turning off a key before changing settings).

## Testing Instructions

This will be easiest to test in conjunction with the corresponding PR in ATEM-state.

## Other Information

A minor version bump and release will be needed as this adds new API surface. It is backwards compatible as a default is provided which won't change behaviour.

VSCode was showing syntax problems for me because the build target for typescript was older than some of the library features being used, so I bumped the version to match `@sofie-automation/code-standard-preset`. I didn't upgrade the dependency on that project because that caused other issues which will need addressing.

## Status

- [x] Relevant documentation (code comments, [system documentation](https://sofie-automation.github.io/sofie-core//)) has been added / updated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   ],
   "scripts": {
     "postinstall": "./node_modules/.bin/husky install",
-    "prepack": "pinst --disable",
+    "prepack": "pinst --disable && run build:main",
     "postpack": "pinst --enable",
     "build": "rimraf dist && run build:main",
     "build:main": "tsc -p tsconfig.build.json",
--- a/src/commands/CommandBase.ts
+++ b/src/commands/CommandBase.ts
@@ -23,12 +23,15 @@ export abstract class DeserializedCommand<T> implements IDeserializedCommand {
 
 export interface ISerializableCommand {
 	serialize(version: ProtocolVersion): Buffer
+	runOrderGroup?: number
 }
 
 /** Base command type for a simple writable command, which has a few values which must all be sent */
 export abstract class BasicWritableCommand<T> implements ISerializableCommand {
 	public static readonly rawName?: string
 	public static readonly minimumVersion?: ProtocolVersion
+
+	public runOrderGroup?: number = 0
 
 	protected _properties: T
 
--- a/src/commands/CommandBase.ts
+++ b/src/commands/CommandBase.ts
@@ -23,6 +23,14 @@ export abstract class DeserializedCommand<T> implements IDeserializedCommand {
 
 export interface ISerializableCommand {
 	serialize(version: ProtocolVersion): Buffer
+	/**
+	 * Controls the command's execution order and batching.
+	 *
+	 * Lower values run **earlier**. Commands with different `runOrderGroup` values
+	 * will not be processed in the same batch.
+	 *
+	 * Defaults to 0. Use a **negative value** to ensure execution before the default group.
+	 */
 	runOrderGroup?: number
 }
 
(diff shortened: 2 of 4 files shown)


jstarpl (changes requested):
Looks good, but test suites need updating to match the new structure before this can be merged.

tsconfig.build.json
```diff
@@ -9,6 +9,7 @@
 			"*": ["./node_modules/*"],
 			"atem-connection": ["./src/index.ts"]
 		},
-		"types": ["node"]
+		"types": ["node"],
+		"target": "es2022", // TODO:RJM: Remove es2022 target when we upgrade @sofie-automation/code-standard-preset from v2 to v3
```
Julusian: I don't think this is safe to do currently, as this library still technically supports node14 which only supports es2020

src/commands/CommandBase.ts
```diff
@@ -23,12 +23,14 @@ export abstract class DeserializedCommand<T> implements IDeserializedCommand {
 
 export interface ISerializableCommand {
 	serialize(version: ProtocolVersion): Buffer
+	runOrderGroup?: number
```
nytamin, line 34: Could you add a comment detailing this property?
Something like
```ts
/** When set, ensures that commands with the lowest runOrderGroup are sent first. (Defaults to 0) */
```

Julusian: Probably include something about methods being sent in separate batches, not fully following the original order of the commands

package.json
```diff
@@ -24,7 +24,8 @@
     }
   ],
   "scripts": {
-    "postinstall": "./node_modules/.bin/husky install",
+	"prepare": "run build:main",
+    "postinstall": "husky install || true",
```
Julusian: This reliably fails for me with `command not found: husky`.

It happens because `husky` is not a direct dependency (only via @sofie-automation/code-standard-preset) so yarn doesn't find it as a valid executable.

I suspect this is why you added the `|| true`, but this just means that husky will never be setup

package.json
```diff
@@ -24,7 +24,8 @@
     }
   ],
   "scripts": {
-    "postinstall": "./node_modules/.bin/husky install",
+	"prepare": "run build:main",
```
Julusian: I'm not 100% convinced by this prepare script. Largely because I don't know the full implications of when it will be run.
As far as I can tell from a quick google, `yarn` will never run it, `npm` will as part of publishing or installing from git.

So this looks like it will solve installing from git for npm users, it won't do anything for yarn users (ie us).
Not enough to argue against it, just worth noting that it is a half solution to installing from git.

However, this might 'break' the npm publishing now (or at least the desired flow). Since setting up trusted publishing, we are using `npm publish` and are intentionally doing the build a few steps earlier, before escalating priviledges to have the publish token. It looks like this will 'break' that flow by rebuilding the code as part of the publish command right where I was trying to avoid it

PR #191, opened 2026-02-23
feat: enhance AtemSocket to handle caching

### LINKED ISSUE #190: Bug Report: High CPU load with 2 M/E Constellation HD
### About me

I hold a M.Sc. in Informatics and work at Fraunhofer AISEC. If my limited time allows for it, I will be happy to help sorting this out!

I came over here from https://github.com/bitfocus/companion-module-bmd-atem/issues/127 and this issue might be related to https://github.com/Sofie-Automation/sofie-atem-connection/issues/96.

However, few things are different:
- We are using a "2 M/E Constellation HD" device.
- We observe this behavior in 2 places: One is mentioned Bitfocus Companion Connection (latest, i.e., 3.20.1), the other is a custom application of mine using this very lib (latest version, i.e., 3.8.2) directly to implement a DIY tally light system, running on a Raspi 4B (4 GB RAM model).

### Observed Behavior

Both devices show constant, high CPU consumption, about 30-40 % on both
- an Intel N100 CPU (NUC) running Bitfocus Companion with Atem Connection.
- an Raspberry Pi 4B 4GB running our custom TS-written app as explained above.

### Expected Behavior

Less idle CPU consumption, I guess? 😉 

### Version

sofie-atem-connection 3.8.2, Bitfocus Compantion with "Blackmagic Design: ATEM" v3.20.1

### Severity / Impact

Not mission-critical, though wasteful and annoying.

milux:
## About the Contributor
This PR is a personal contribution to the `sofie-atem-connection` library in order to improve overall efficiency.
Fixes #190.

## Type of Contribution

This is a:

Bug fix / Feature

## Current Behavior
As discussed in #190, highly frequent command batches of `TimeCommand`s followed by other redundant commands with identical contents cause significant CPU load, even when ATEM devices are not actively "doing" anything in particular.

## New Behavior
The CPU load was (again) cut down approximately by half for affected ATEM devices.
This is achieved by two means primarily:
- A full copy of the incoming UDP packet's serialized commands was substituted by a "normalization function" which creates a view on the passed memory instead of copying it.
- The serialized representation of command batches starting with a `TimeCommand` are (partially, after the `TimeCommand` itself) cached. Iff the next batch also starts with a time command **and** otherwise equals the previous batch (byte-wise) entirely. Only the `TimeCommand` is emitted and the remaining contents are dropped without (redundant) deserialization and further processing.

## Testing Instructions
I tried to provide full (jest) test coverage for all added and modified code snippets.

I took precautions to make sure my caching strategy is sound by always flushing/updating the cached copy if a command batch does either not start with a `TimeCommand` or differs by content or length after the `TimeCommand`.

My optimization should be valid **iff** the following assumptions hold:
- Processing of commands is idempotent, i.e., processing the same batches of commands multiple times always yields the same state.
- Any message changing internal state managed by command messages passes through `_parseCommands`, i.e., there is no "side channel" that may cause state changes that would normally be overwritten by the commands received.

If I am mistaken about any of those assumptions, my PR may introduce subtile semantic errors and **should not be merged** as is.

## Other Information
Since this is my first contribution to this lib, I want to kindly ask you carefully reviewing my changes and see if they fit your expectations.
Specifically, please see if my assumptions above hold. If they do, I am optimistic that this PR can be safely merged.

## Status

- [x] Relevant documentation (code comments, [system documentation](https://sofie-automation.github.io/sofie-core//)) has been added / updated.

## Overview

Implements an optimization to reduce CPU load from frequent, redundant ATEM command batches (notably TimeCommand-heavy traffic). Adds payload normalization to avoid buffer copies and a caching mechanism to skip redundant deserialization of command batch remainders when batches start with a TimeCommand.

### Core Implementation (src/lib/atemSocket.ts)
- Exports new type: `ThreadedPayload = Buffer | Uint8Array | { type: 'Buffer'; data: number[] }`.
- Payload normalization:
	- Adds private `_normalizePayload(payload: ThreadedPayload)` to convert supported payload variants to a native `Buffer` view without making full copies; returns `undefined` for invalid shapes.
	- Applies normalization in the threaded-socket onCommandsReceived callback before parsing.
- TimeCommand remainder caching:
	- Adds private `_lastTimeCommandRemainder: Buffer | undefined` to cache the bytes following a leading `TimeCommand`.
	- `_parseCommands()` tracks the first command; if it is a `TimeCommand` the remainder is compared against the cached remainder. If the remainder equals the cache, only the `TimeCommand` is processed and parsing stops. If different, the remainder is cached via `Buffer.from(remainder)`. If the first command is not a `TimeCommand`, the cache is cleared.
	- Adds stricter length validation for commands (`length < 8 || length > buffer.length`).
- Connection lifecycle:
	- Resets `_lastTimeCommandRemainder` on connect and disconnect.
- Error/debug behavior preserved for invalid payloads and deserialization failures.

### Public API / Type Changes
- `ThreadedPayload` exported from the AtemSocket module.
- Threaded socket child callback signature updated to accept `ThreadedPayload`; tests and mocks updated accordingly.

### Tests (src/lib/__tests__/atemSocket.spec.ts)
- Tests adapted to use `ThreadedPayload` and the normalization path.
- New/updated tests cover:
	- TimeCommand remainder deduplication and cache-update/clear behavior.
	- Payload normalization variants (`Uint8Array`, `{ type: 'Buffer'; data: number[] }`) and invalid payload handling.
- Mocked AtemSocketChild and ThreadedClassManager usage adjusted to the new callback types.

## Implications & Assumptions
- Correctness depends on command processing being idempotent and all state-changing messages flowing through `_parseCommands()` (author-documented assumptions). Reviewers should validate these assumptions before merging.
- Cache semantics: only applies when the first command is a `TimeCommand`; flushed whenever a batch starts with a different command or the remainder differs in content or length.

## Performance Impact
- Reduces memory allocations by creating Buffer views instead of full copies and avoids redundant deserialization/processing for repeated TimeCommand-led batches; the author reports roughly a 50% CPU reduction in affected scenarios.

### DIFF
--- a/src/lib/__tests__/atemSocket.spec.ts
+++ b/src/lib/__tests__/atemSocket.spec.ts
@@ -3,6 +3,7 @@
 import {
 	CutCommand,
 	ProductIdentifierCommand,
+	TimeCommand,
 	VersionCommand,
 	ProgramInputUpdateCommand,
 	PreviewInputUpdateCommand,
@@ -11,7 +12,7 @@ import {
 	DeserializedCommand,
 } from '../../commands'
 import { ProtocolVersion, Model } from '../../enums'
-import { AtemSocket } from '../atemSocket'
+import { AtemSocket, ThreadedPayload } from '../atemSocket'
 import { ThreadedClass, ThreadedClassManager } from 'threadedclass'
 import { Buffer } from 'buffer'
 import { CommandParser } from '../atemCommandParser'
@@ -24,7 +25,7 @@ jest.mock('../atemSocketChild')
 class AtemSocketChildMock implements AtemSocketChild {
 	public onDisconnect: () => Promise<void>
 	public onLog: (message: string) => Promise<void>
-	public onCommandsReceived: (payload: Buffer, packetId: number) => Promise<void>
+	public onCommandsReceived: (payload: ThreadedPayload, packetId: number) => Promise<void>
 	public onPacketsAcknowledged: (ids: Array<{ packetId: number; trackingId: number }>) => Promise<void>
 
 	constructor() {
@@ -49,7 +50,7 @@ const AtemSocketChildSingleton = new AtemSocketChildMock()
 		_opts: any,
 		onDisconnect: () => Promise<void>,
 		onLog: (message: string) => Promise<void>,
-		onCommandsReceived: (payload: Buffer, packetId: number) => Promise<void>,
+		onCommandsReceived: (payload: ThreadedPayload, packetId: number) => Promise<void>,
 		onPacketsAcknowledged: (ids: Array<{ packetId: number; trackingId: number }>) => Promise<void>
 	) => {
 		AtemSocketChildSingleton.onDisconnect = onDisconnect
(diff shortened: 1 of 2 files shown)


milux:
@coderabbitai review

milux:
@Julusian CodeRabbit was initially unhappy with my cache handling in some edge cases. Had some more iterations but think it is very robust now.
Would be great if you could have a look at it.

Julusian:
could you send me an xml export of your atem? Im not sure what I need to change to reproduce this, the old trigger doesnt appear to cause this anymore.

I'm currently wondering whether the risk (and minor complexity) this introduces is worth the performance gain, vs potentially finding another solution. I'm a little worried about the potential for causing something to be lost if it gets dumped at the end of the sequence. Maybe a bit cautious, I can't think of what could be there that could happen in 2 frames in a row. 

I think this will be invalidated unnecessarily at times (eg, when theres a cut, that gets put after the TimeCommand in the same packet), but thats fine, as itll only be 2 packets of that.
When fairlight audio levels are being streamed, the TimeCommand appears to move to be in the middle/end of a packet, which will cause it to bypass this cache, but thats fine too and probably negligable compared to the cost of the rest of the packets.

I'm also wondering how much actual performance gain this gives, as parsing those commands cant be all that much work surely. its a bit of extra parsing of the buffers and object construction, but [we are swallowing the change](https://github.com/Sofie-Automation/sofie-atem-connection/blob/1b18fd40aa535a9b0b9447cec852e8278253bf45/src/commands/Settings/MultiViewerSourceCommand.ts#L70) when it is doing nothing, so it shouldnt be leaving the bounds of the library code.  
So I would like to understand where the cost lies

milux:
> could you send me an xml export of your atem? Im not sure what I need to change to reproduce this, the old trigger doesnt appear to cause this anymore.

Funny, I just realized from the Source Code you referenced below that this issue on our device series has been known (and documented plus partially mitigated by you) since 2023. :laughing: 

I created a XML export and can send it via any channel of your choice, but I'm 99 % sure that it won't provide you any helpful insight.

> I'm currently wondering whether the risk (and minor complexity) this introduces is worth the performance gain, vs potentially finding another solution. I'm a little worried about the potential for causing something to be lost if it gets dumped at the end of the sequence. Maybe a bit cautious, I can't think of what could be there that could happen in 2 frames in a row.
>
> I think this will be invalidated unnecessarily at times (eg, when theres a cut, that gets put after the TimeCommand in the same packet), but thats fine, as itll only be 2 packets of that. When fairlight audio levels are being streamed, the TimeCommand appears to move to be in the middle/end of a packet, which will cause it to bypass this cache, but thats fine too and probably negligable compared to the cost of the rest of the packets.

Right, it will be invalidated unnecessarily often times because I implemented it as defensive as possible to get the risk of error down to the absolute minimum.

As stated, the cache is flushed iff (equivalence relation!) the sequence does not start with a `TimeCommand` or otherwise differs in any byte from the previous batch/frame in any way. This is very effective on the buggy Constellation HD devices and costs basically nothing since the cached buffer "is already there".

The **one and only risk I cannot safely rule out** is "external state mutation" between frames with identical data, i.e., whether there is some code in the lib that could cause a state change which is not reflected by a command from the ATEM right after and might remain stale due to skipping of cached commands.

> I'm also wondering how much actual performance gain this gives, as parsing those commands cant be all that much work surely. its a bit of extra parsing of the buffers and object construction, but [we are swallowing the change](https://github.com/Sofie-Automation/sofie-atem-connection/blob/1b18fd40aa535a9b0b9447cec852e8278253bf45/src/commands/Settings/MultiViewerSourceCommand.ts#L70) when it is doing nothing, so it shouldnt be leaving the bounds of the library code. So I would like to understand where the cost lies

I totally agree with you in every part but for the "as parsing those commands cant be all that much work surely" assumption.

TBH, I thought so too, until I did the profiling and noticed that even this parsing and checking of 32 unspectacular `MultiViewerSourceUpdateCommand` instances against the current state is causing significant load when run 50 times a second.

I have to speculate about precise reasons here, but I assume that something of the comparison logic might work against caches or the prediction units of the CPU. Further, I have observed heavy fluctuation of the heap utilization before my caching patch which almost entirely vanished. That actually makes perfect sense since about 1.6k `MultiViewerSourceUpdateCommand` objects per second were flooding the heap previously.

Julusian:
I haven't entirely forgotten about this, but I'm still not 100% sure on if it introduces risk of error or not. With the range of users, that makes me nervous.

Maybe this could be gated behind a boolean on `AtemOptions`? That way it will be opt-in and it can get some real world testing before deciding to enable it by default in a future version.

Or, I know we don't have the structure in place, but maybe this should be gated on models that we know do or could have this issue?

milux:
> I haven't entirely forgotten about this, but I'm still not 100% sure on if it introduces risk of error or not. With the range of users, that makes me nervous.
> 
> Maybe this could be gated behind a boolean on `AtemOptions`? That way it will be opt-in and it can get some real world testing before deciding to enable it by default in a future version.
> 
> Or, I know we don't have the structure in place, but maybe this should be gated on models that we know do or could have this issue?

Both approaches sound reasonable to me.
Thank you for keeping that on your radar.
The last benchmarks on better hardware didn't show the same effect strength anyway, but I would still appreciate this being included.

If you want me to implement either one of that approaches, please tell me what I shall do.
If you have some time to quickly code the flag or device-dependent behavior yourself - even better. :)

PR #202, opened 2026-08-14
fix: parse SuperSource border via reverse-engineered SSSB command

bartcuijpers:
## Problem

On some ATEM units, `SuperSource.border` in state stays permanently `undefined`. Confirmed on an **ATEM 4 M/E Constellation HD** (model 20, `apiVersion` 0x00020020 / `ProtocolVersion.V9_6`): the unit never sends the documented `SSBd` (`SuperSourceBorderUpdateCommand`) packet - not on connect, and not even in response to a `CSBd` write.

## Investigation

Live packet capture (see raw command dump in the linked conversation, happy to share details) showed this unit instead emits an undocumented `SSSB` command, once per SuperSource box, whenever a border setting is changed via the panel or ATEM Software Control. This command had no registered parser in the library and was previously silently dropped as `"Unknown command"` in `atemSocket.ts`.

By diffing raw `SSSB` payloads against isolated, single-control UI changes (toggling border on/off; nudging hue/saturation/luma/outer width/inner width up and down by known amounts; dragging outer/inner width sliders to their documented min (0.00) and max (16.00)), I was able to reverse engineer and calibrate the packet layout with fairly high confidence - see the extensive comment in `SuperSourceBoxBorderCommand.ts`.

Notably:
- `borderEnabled`, `borderHue`, `borderSaturation`, `borderLuma` map directly (no rescaling needed - values already fall in the same ranges as the documented `SSBd` fields, and the observed default luma of `1000` matches the known default exactly).
- `borderOuterWidth`/`borderInnerWidth` are transmitted using the packet's full `0-65535` range for `0.00-16.00`, rather than the `0-1600` (hundredths) convention used by `SSBd`/`CSBd` elsewhere in this library. This PR rescales them to that convention on parse.
- Bevel, bevel softness/position, outer/inner softness, and light source direction/altitude are **not present** in this packet - those controls are greyed out in ATEM Software Control for this unit, suggesting the hardware doesn't support them at all. These fields are left untouched (kept from previously known state, or a neutral default) rather than guessed at.
- Separately (not addressed by this PR): `setSuperSourceBorder()` writes are acked by this unit but appear to have **no effect at all** - a write-path issue independent of this read-side fix.

## ⚠️ Caveat - please read

This command is **reverse-engineered against a single hardware unit** and is not part of any official/documented ATEM protocol reference I have access to. It's entirely possible other SuperSource-capable models either don't send `SSSB` at all (harmless - this parser simply won't be exercised), or send it with a **different byte layout** (which could produce incorrect `border` values on those models).

I'd really appreciate verification/testing on other SuperSource-capable hardware before this is considered broadly safe. Happy to gate this behind something more conservative if maintainers prefer (e.g. only apply if `SSBd` has never been seen for that SuperSource, or similar), or to hear if anyone has an official reference for this command's real layout/name.

## Testing

- Added a dedicated unit test (`SuperSourceBoxBorderCommand.spec.ts`) covering parsing, width rescaling, and `applyToState` merge behaviour, since there's no `libatem-data.json` fixture for this undocumented command (excluded it from the "all commands have test cases" check, same precedent as the existing multiviewer border exclusions).
- Verified live against the reference unit: `video.superSources.N.border` now populates correctly and tracks live changes made via ATEM Software Control.
- Full existing test suite (1909 tests), lint, and typecheck all pass unchanged.


Julusian:
Thanks for digging into this. Ill try and have a proper look soon and investigate a bit myself (I have a 2ME constellation HD). Im surprised noone has noticed this yet tbh..

Not to worry about reverse engineering, that is how all of this library is written. the libatem stuff is a c# library I started prior to this which has some better integration tests against the c++ sdk, so is being used as a point of reference when possible. But I've gotten very bad at keeping that up to date.

```
 * 24 byte payload layout, as reverse engineered by diffing packets against
 * live UI changes in ATEM Software Control:
 *   0:     ssrcId (u8)
 *   1:     boxId (u8)
 *   2:     borderEnabled (u8, 0/1)
 *   3:     unknown, box-specific, unaffected by border changes
 *   4-7:   outerWidth, duplicated across two u16 BE values
 *   8-15:  innerWidth, duplicated across four u16 BE values
 *   16-17: hue (u16 BE)
 *   18-19: saturation (u16 BE)
 *   20-21: luma (u16 BE, default 1000 - matches the known SSBd luma scale)
 *   22-23: unknown, box-specific, unaffected by border changes
 ```

A few comments on this:
* Byte 1: a boxId is interesting to see here. That suggests the border of each box can be set independently. I think we should explore this, as if that is true then we will want to put this in a new place in the state where it can be per box.
* Byte 3: yeah this will be padding. 16bit values are always aligned to 2byte offsets 32bit values to 4byte offsets.
* Byte 22-23: Likely also just padding, commands are sized in multiples of 4bytes.
Both those padding will contain random bytes that are meaningless, so can be ignored

Have you checked to see if the atem software sends a different command when setting the border properties? The atem software in recent versions hasnt exposed everything the atem can do (eg audio routing flexibility), so it could be that there is something lurking in the SDK for this

No need to gate anything here (so far at least), as this will only react when it receives one of these from the atem, so on other models (or older firmware?) it wont ever be reached

Julusian:
Taking a look at this, and I can reproduce this on the 2ME HD.
Looking into it a bit, when originally released, this model lacked any borders on the supersource. According to the atem software control ui, there is still no border support. But in 9.6.0 (august 2024), the firmware added support for per-box borders. I have not been keeping up with my attempt to reverse engineer the full protocol, so didnt notice the addition


There is a corresponding CSSB command to set these properties. 

My first working demo (notice the blue box2 which is not being affected):
https://github.com/user-attachments/assets/b0cb4960-f3e8-4cce-a11f-e3b5d44bb85b

I think everything in this branch is now correct

